PR #15, opened 2021-09-03
fixed misleading error message when S3_SECRET_KEY is not present

vinodpandey:


dekobon:
This looks good. Thanks for the contribution!

PR #18, opened 2022-02-08
Use temporary credentials to access S3 bucket

### LINKED ISSUE #8: Use Instance profile credential
Is it possible to use temporary credential retrieved from instance profile/role?

FlorianSW:
This PR adds two new possible paths from where credentials can be retrieved:
- the ECS credentials endpoint, when running in an ECS task
- IMDSv2 role credentials endpoint, when running on an EC2 instance

The currently used static credentials from the environment still works as well.
This code will try to automatically select the best approach of getting credentials in the following order
(the first ones are favoured over the last ones):
- Static credentials in the documented `S3_` environment variables
- ECS credentials endpoint (when the relative credentials URI environment variable is set, which is done by AWS ECS)
- fallback to IMDSv2 (if IMDS is not available, the container will not start)

Fixes #8 

Example template of how to deploy to ECS:
[template.yml.zip](https://github.com/nginxinc/nginx-s3-gateway/files/8026416/template.yml.zip)


dekobon:
Thank you so much for this contribution. As I dive into the code review, based on a quick glance of the PR, there are three items that I would like to include in this change. I can do my best to fill them in or if you (@FlorianSW) can add them it would be much appreciated.

1. Unhardcode references to `169.254.169.254`. I know this may be overkill, but at the least by putting it in a variable it is easier to tell the story about what that IP is for to someone reading the code.
2. Provide a quickstart guide to setting up temporary credentials with ec2.
3. Provide a quickstart guide to setting up temporary credentials with ecs.

FlorianSW:
@dekobon You're welcome to change whatever you think is needed :) 

Regarding the reference to `169.254.169.254`: Do you mean making it _configurable_ or "just" extracting the IP to a variable in the script (in order to make the code a bit more readable)? I would +1 the second part, but for the first one, I would vote against it. This IP is static for that specific use case. It doesn't make much sense to make it configurable, without knowing why someone would even want to do that.

To the setup with EC2: That should be the same as with any other VM setup. The EC2 instance should only get a role associated to it. How to do that is explained in the AWS docs: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#launch-instance-with-role I never really worked with EC2 instance (apart from a bit here and there), so I don't really know much about that without doing it once. But it should be straight forward tbh.

For the ECS part: I thought about that already and wanted to add the CloudFormation template I used to test the ECS integration. I obviously forgot about that in the PR 🤣 Let me add the template to the original PR text. With it, it is as easy as pushing a built image to an ECR and creating a stack with it. I'm not sure if that should be part of the repo?

dekobon:
Regarding item 1, yeah I was thinking primarily for readability. If it ever needs configuring we can work on that when the time comes.
As for items 2 and 3, I'm asking for primarily selfish reasons. I don't know how to set up so that I can test it. I'm going through the AWS docs, but I'm a slow learner. I suspect that others may also struggle, so at least having some notes about how to configure it may be useful.

FlorianSW:
> Regarding item 1, yeah I was thinking primarily for readability. If it ever needs configuring we can work on that when the time comes.

Got it, let me change that and add a new commit to the PR 👍 

> As for items 2 and 3, I'm asking for primarily selfish reasons. I don't know how to set up so that I can test it. I'm going through the AWS docs, but I'm a slow learner. I suspect that others may also struggle, so at least having some notes about how to configure it may be useful.

I added the CloudFormation template I used to test the ECS feature (in a zip, as GH does not support to add yml directly...). This will setup all the required things like an ECS service + task and a load balancer (which is maybe not needed if you attach a public IP to the task itself, so there might be room to simplify this stack massively).

For the EC2 part, I don't have the possibility right now to check that out, really. I can only imagine what to do, but not actually try it out :/ If others are using EC2 instances, that would be a massive help here (testing and maybe even an example CloudFormation template or something).

dekobon:
I just reviewed the [code for the stand-alone systemd example install](https://github.com/nginxinc/nginx-s3-gateway/blob/master/examples/ubuntu_install/ubuntu_setup.sh) and it looks like that will need to be updated as well. I'm think we will make a separate ticket for it and do it post-merge.

FlorianSW:
> I just reviewed the [code for the stand-alone systemd example install](https://github.com/nginxinc/nginx-s3-gateway/blob/master/examples/ubuntu_install/ubuntu_setup.sh) and it looks like that will need to be updated as well. I'm thinking we will make a separate ticket for it and do it post-merge.

Fixed that in the last commit, as well as the extraction of the IP addresses 👍

common/etc/nginx/include/s3gateway.js
```diff
@@ -119,6 +132,15 @@ function awsHeaderDate(r) {
     return _amzDatetime(now, _eightDigitDate(now));
 }
 
+function readCredentials() {
+    try {
+        var creds = fs.readFileSync('/tmp/credentials.json');
```
dekobon, line 137: Unit tests fail when the file is `/tmp/credentials.json` does not exist (which would be the default state in a unit test run). We need to figure out a way to mock the file read or to add a file with fake values within the execution of the tests.

test/unit/s3gateway_test.js
```diff
@@ -265,7 +267,72 @@ function testEscapeURIPathPreservesDoubleSlashes() {
...
+                });
+            } else {
+                throw 'Invalid token passed: ' + options.headers['x-aws-ec2-metadata-token'];
+            }
+        } else {
+            throw 'Invalid request URL: ' + url;
+        }
+    };
+
+    await s3gateway.fetchCredentials();
```
dekobon, line 328: A mock request object is not passed to the `s3gateway.fetchCredentials()` like in other tests. This typically looks like:
```javascript
    var r = {
        "headersOut" : {
            "Accept-Ranges": "bytes",
            "Content-Length": 42,
            "Content-Security-Policy": "block-all-mixed-content",
            "Content-Type": "text/plain",
            "X-Amz-Bucket-Region": "us-east-1",
            "X-Amz-Request-Id": "166539E18A46500A",
            "X-Xss-Protection": "1; mode=block"
        }
    }

    r.log = function(msg) {
        console.log(msg);
    }
```
This causes a unit test failure in this function (which is called s3gateway.js:697):
```javascript
function _debug_log(r, msg) {
    if (debug && "log" in r) {
        r.log(msg);
    }
}
```

common/etc/nginx/include/s3gateway.js
```diff
@@ -643,10 +683,108 @@ function _require_env_var(envVarName) {
...
+        },
+        method: 'PUT',
+    });
+    var token = await tokenResp.text();
+    var resp = await ngx.fetch(ec2ImdsSecurityCredentialsEndpoint, {
+        headers: {
+            'x-aws-ec2-metadata-token': token,
+        },
+    });
+    // TODO: I'm lazy, but this can be a list of roles, however, for sake of simplicity, let's assume it's just one
```
dekobon, line 762: Let's change this TODO to an explanation that we are only doing a single role right now and then let's create a ticket for it.

common/etc/nginx/include/s3gateway.js
```diff
@@ -643,10 +683,108 @@ function _require_env_var(envVarName) {
     }
 }
 
+var maxValidityOffset = 270000;
```
dekobon, line 686: It would be good to add a comment or change the variable name here to indicate the unit of measurement for the time offset interval.

dekobon (changes requested):
This is awesome work! I'm really excited to see this get integrated.

common/etc/nginx/include/s3gateway.js
```diff
@@ -119,6 +118,15 @@ function awsHeaderDate(r) {
     return _amzDatetime(now, _eightDigitDate(now));
 }
 
+function readCredentials() {
+    try {
+        var creds = fs.readFileSync('/tmp/credentials.json');
```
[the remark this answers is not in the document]
FlorianSW: Done.

common/etc/nginx/include/s3gateway.js
FlorianSW: Hmm, this file being missing shouldn't be any problem, really. I saw the failing test here:
https://github.com/nginxinc/nginx-s3-gateway/runs/5117512199?check_suite_focus=true
However, looking at it doesn't give me any clue why they are failing. Do you have an exact line number in the logs where the reason for the failing test is given?

I don't think that mocking the file is a good idea from the test perspective. The test shouldn't care what temporary storage the implementation is using to cache credentials. It could easily change to environment variables at some point. The test should then not need to change just because we might change the implementation. I think it is better to find the issue why the code can not work without this file being present (in my tests in ECS, it worked pretty well and the file was not there on the first request, either).

test/unit/s3gateway_test.js
FlorianSW: 👍

common/etc/nginx/include/s3gateway.js
FlorianSW: I looked it up when doing this code, the thing is:
- the AWS SDK in Go, iirc, reads each line as a role and selects the first one by default.
- an EC2 instance _can_ only have one role attached

Therefore, I think it might be just a safeguard in the SDK, as maybe the IMDS can return multiple roles. However, in the current scenario (EC2 instances) there can only be one role at any time. I change the comment to reflect that, though 👍

common/etc/nginx/include/s3gateway.js
FlorianSW: 👍

common/etc/nginx/include/s3gateway.js
FlorianSW: Meaning also: I don't think we need a separate issue for that, tbh. There's currently no use-case, I know, which uses IMDS with multiple roles.

test/unit/s3gateway_test.js
```diff
@@ -265,7 +267,100 @@ function testEscapeURIPathPreservesDoubleSlashes() {
...
+    process.env['S3_ACCESS_KEY_ID'] = undefined;
+    process.env['AWS_CONTAINER_CREDENTIALS_ABSOLUTE_URI'] = undefined;
+    globalThis.ngx.fetch = function (url, options) {
+        if (url === 'http://169.254.169.254/latest/api/token' && options && options.method === 'PUT') {
+            return Promise.resolve({
+                text: function () {
+                    return Promise.resolve('A_TOKEN');
+                },
+            });
+        } else if (url === 'http://169.254.169.254/latest/meta-data/iam/security-credentials/s3access') {
```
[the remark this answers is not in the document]
FlorianSW: Ah, good catch! That was a copy-paste from one of my first commits. This line basically only works, if the role attached to the EC2 instance is named "s3access", which is usually not the case. I'll change that 👍

test/unit/s3gateway_test.js
```diff
@@ -265,7 +267,100 @@ function testEscapeURIPathPreservesDoubleSlashes() {
...
+            "Content-Type": "text/plain",
+            "X-Amz-Bucket-Region": "us-east-1",
+            "X-Amz-Request-Id": "166539E18A46500A",
+            "X-Xss-Protection": "1; mode=block"
+        }
+    };
+    r.log = function(msg) {
+        console.log(msg);
+    }
+
```
dekobon, line 355: Here we will need to add an additional function to the `r` instance when running as a unit test in order to avoid an exception (which is what is causing the segfault - a patch will be included in the next njs release):
```javascript
    r.return = function(code, body) {
        console.log(`RESPONSE RETURN: ${code} ${body}`);
    }
```

common/etc/nginx/include/s3gateway.js
```diff
@@ -643,10 +690,120 @@ function _require_env_var(envVarName) {
...
+    _debug_log(r, 'Cached credentials are expired or not present, requesting new ones');
+
+    var credentials;
+    if (process.env['S3_ACCESS_KEY_ID']) {
+        credentials = {
+            accessKeyId: process.env['S3_ACCESS_KEY_ID'],
+            secretAccessKey: process.env['S3_SECRET_KEY'],
+            sessionToken: null,
+        };
+    } else if (process.env['AWS_CONTAINER_CREDENTIALS_RELATIVE_URI']) {
```
dekobon, line 723: The simple existence test here is not sufficient as it fails to match in unit tests. Perhaps we should do something like:
```javascript
} else if ('AWS_CONTAINER_CREDENTIALS_RELATIVE_URI' in process.env) {
```

common/etc/nginx/include/s3gateway.js
```diff
@@ -643,10 +690,120 @@ function _require_env_var(envVarName) {
...
+        var exp = new Date(current.expiration).getTime() - maxValidityOffsetMs;
+        if (now.getTime() < exp) {
+            r.return(200);
+            return;
+        }
+    }
+    _debug_log(r, 'Cached credentials are expired or not present, requesting new ones');
+
+    var credentials;
+    if (process.env['S3_ACCESS_KEY_ID']) {
```
dekobon, line 717: The simple existence test here is not sufficient as it fails to match in unit tests. Perhaps we should do something like:
```javascript
} else if ('S3_ACCESS_KEY_ID' in process.env) {
```

test/unit/s3gateway_test.js
FlorianSW: I will add the function, but I don't think it is valuable to log the whole response, isn't it? I might simply add an "assertion" that the correct code is returned, ok? :)

common/etc/nginx/include/s3gateway.js
FlorianSW: Hmm, you sure? I ran the tests once (it is a bit difficult for me to run the tests except in the CI tbh :(), and they worked. During the patches, I changed, however, the env variable back to the one injected by AWS (AWS_CONTAINER_CREDENTIALS_RELATIVE_URI), the unit tests still sets the non-default one (AWS_CONTAINER_CREDENTIALS_ABSOLUTE_URI), I'll change that and then this should work (at least on my newbie understanding, you might still be right 🙈)

examples/ubuntu_install/ubuntu_setup.sh
```diff
@@ -28,59 +28,7 @@ if ! dpkg --status wget 2>/dev/null | grep --quiet Status > /dev/null; then
...
-
-if [ -z "${S3_STYLE}" ]; then
-  >&2 echo "S3_STYLE must be set"
-  exit 1
-fi
-
-if [ -z "${S3_DEBUG}" ]; then
-  S3_DEBUG=false
-fi
+source ../../common/check_env.sh
```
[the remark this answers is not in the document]
FlorianSW: Am I missing something in the diff? I don't see where the parts I touched modifies env variables 🤔

common/etc/nginx/include/s3gateway.js
dekobon: Sadly, I'm not *that* sure. I did some experiments in njs with trying to add in new values to `process.env` and found that I couldn't do it. How did you run the unit tests? I'm running them by executing:
```
/usr/bin/njs -t module -p common/etc/nginx test/unit/s3gateway_test.js
```

If you run `test.sh` it will also execute the unit tests in addition to the integration tests.

examples/ubuntu_install/ubuntu_setup.sh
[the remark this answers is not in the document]
dekobon: My apologies, I was trying to say that there is a fair bit more work to do here in addition to the changes already done.

test/unit/s3gateway_test.js
dekobon: Great! That dealt with one of the two segfaults that were happening. Using an assertion here makes a lot of sense.

common/docker-entrypoint.d/00-check-for-required-env.sh
```diff
@@ -20,71 +20,7 @@
...
-
-if [ "${AWS_SIGS_VERSION}" != "2" ] && [ "${AWS_SIGS_VERSION}" != "4" ]; then
-  >&2 echo "AWS_SIGS_VERSION contains an invalid value (${AWS_SIGS_VERSION}). Valid values: 2, 4"
-  failed=1
-fi
-
-if [ $failed -gt 0 ]; then
-  exit 1
-fi
+source ../check_env.sh
```
dekobon, line 23: This shell script uses `sh`, so it doesn't properly support the `source` bash keyword. Here we should use `. ../check_env.sh` instead.

common/etc/nginx/include/s3gateway.js
dekobon: I did another simpler test:

```javascript
process.env['FOO'] = 'bar';
console.log(process.env['FOO']);
```

```
$ njs process_env.js 
bar
```

In this case it looks like it is successfully modifying the environment variables. I'm going to dive in a bit deeper.

common/etc/nginx/include/s3gateway.js
```diff
@@ -643,10 +690,120 @@ function _require_env_var(envVarName) {
...
+    var token = await tokenResp.text();
+    var resp = await ngx.fetch(ec2ImdsSecurityCredentialsEndpoint, {
+        headers: {
+            'x-aws-ec2-metadata-token': token,
+        },
+    });
+    // This _might_ get multiple possible roles in other scenarios, however, EC2 supports attaching one role only.
+    // It should therefore be safe to take the whole output, even given IMDS _might_ (?) be able to return multiple
+    // roles.
+    var credName = resp.text();
```
dekobon, line 782: Shouldn't we be doing an `await` here? Like:
```
var credName = await resp.text();
```

common/etc/nginx/templates/default.conf.template
```diff
@@ -78,6 +90,7 @@ server {
 
         # Set the Authorization header to the AWS Signatures credentials
         proxy_set_header Authorization $s3auth;
+        proxy_set_header X-Amz-Security-Token $s3SecurityToken;
```
dekobon, line 93: We will also need to add this line to the `location @s3Listing` block as well. Without it, directory listings won't work.

dekobon:
I just discovered that in order to use the AWS metadata endpoint from within a container running in EC2, you need to run the following command to allow for additional network hops from your EC2 instance:
```
aws ec2 modify-instance-metadata-options  
--instance-id <instance id> --http-put-response-hop-limit 3 --http-endpoint enabled
```
We will want to add a note about that in the documentation.

common/etc/nginx/include/s3gateway.js
```diff
@@ -643,10 +690,120 @@ function _require_env_var(envVarName) {
...
+    if (credName === "") {
+        throw 'No credentials available for EC2 instance';
+    }
+    resp = await ngx.fetch(ec2ImdsSecurityCredentialsEndpoint + credName, {
+        headers: {
+            'x-aws-ec2-metadata-token': token,
+        },
+    });
+    var creds = resp.json();
```
dekobon, line 791: Shouldn't we be doing an `await` there? Like:
```javascript
var creds = await resp.json();
```

common/etc/nginx/include/s3gateway.js
FlorianSW: I used test.sh. Let me try to run them again (in both ways). However, I use a windows environment (together with cygwin) 🤦 which made things a bit harder iirc.

common/etc/nginx/include/s3gateway.js
FlorianSW: Embarrassing, you're absolutely right, my fault 👍

common/etc/nginx/templates/default.conf.template
FlorianSW: Good catch! I didn't try out directory listing so far 🙈

common/etc/nginx/include/s3gateway.js
FlorianSW: yep, same as above 🙈

FlorianSW:
> I just discovered that in order to use the AWS metadata endpoint from within a container running in EC2, you need to run the following command to allow for additional network hops from your EC2 instance:
> 
> ```
> aws ec2 modify-instance-metadata-options  
> --instance-id <instance id> --http-put-response-hop-limit 3 --http-endpoint enabled
> ```
> 
> We will want to add a note about that in the documentation.

That is needed when you run a docker daemon on the EC2 instance on your own and run nginx with this scripts there, right? I would expect that this is not happening when running ECS with the EC2 launch type (instead of Fargate, which I'm used to use), isn't it? In this case, it should use the ECS container credentials URL.

We should generally add some kind of documentation for this feature, where such a hint would be best suited, yes :)

common/etc/nginx/include/s3gateway.js
FlorianSW: I just run the tests (after I fixed some more issues with mocking a Response promise 🙈) and they run fine now (the unit tests, in a docker container) :)

dekobon:
@FlorianSW I think the code is a good state. I'd like to make a feature branch and merge the changes into it. There are some changes I would like to do. Would you be like to be a reviewer?

FlorianSW:
> @FlorianSW I think the code is a good state. I'd like to make a feature branch and merge the changes into it. There are some changes I would like to do. Would you be like to be a reviewer?

Of course, that's absolutely fine 👍 sure I'll review your changes and comment or built up on them if needed.

dekobon:
I'd like to move discussion about instance profile implementation to github discussions: https://github.com/nginxinc/nginx-s3-gateway/discussions/24

dekobon:
A few updates on the current progress:
* A working version is in the branch [issue/8](https://github.com/nginxinc/nginx-s3-gateway/tree/issue/8).
* We are pushing builds from this branch to the github container repo with the tag `latest-njs-oss`.
* The next item to do is to test ECS deployments.
* Following that, we need to add documentation for running with instance profile credentials.
* Lastly, we will not be able to merge until the next version of njs is released that fixes a bug that we hit with the instance profile credential code path.

All in all, we are pretty close. If anyone has tips on deploying with ECS, it would be much appreciated.

FlorianSW:
Sorry for not being that active lately 🙈 

> * Following that, we need to add documentation for running with instance profile credentials.

I'll add some docs and push it to this PR (also updating to issue/8).

> * Lastly, we will not be able to merge until the next version of njs is released that fixes a bug that we hit with the instance profile credential code path.

What is this about? 🤔

ajschmidt8:
> If anyone has tips on deploying with ECS, it would be much appreciated.

I have an AWS CloudFormation file that I can adapt and share for purposes of testing. I will post it shortly!

ajschmidt8:
Ok, here is the CloudFormation file and usage instructions. Hopefully this helps with testing. Let me know if you guys run into any issues with this.

<details>
<summary>CloudFormation File</summary>

```yaml
AWSTemplateFormatVersion: '2010-09-09'
Parameters:
  NewBucketName:
    Default: <ENTER_RANDOM_BUCKET_NAME_HERE>
    Description: S3 Bucket Name
    Type: String
  Subnet1:
    Default: <ENTER_YOUR_VALUE_HERE>
    Description: ID of the first subnet to be used for resources
    Type: String
  Subnet2:
    Default: <ENTER_YOUR_VALUE_HERE>
    Description: ID of the second subnet to be used for resources
    Type: String
  VpcId:
    Default: <ENTER_YOUR_VALUE_HERE>
    Description: ID of the VPC to be used for resources
    Type: String
  ContainerName:
    Default: s3gateway
    Description: Name of the NGINX Container. No need to change this
    Type: String
  ResourceNamePrefix:
    Default: nginx-s3-gateway
    Description: Common prefix used for resource names. No need to change this
    Type: String
Outputs:
  PublicDNS:
    Description: DNS name of load balancer
    Value: !GetAtt 'ALB.DNSName'
Resources:
  ALB:
    Properties:
      Name: !Join
        - '-'
        - - !Ref 'ResourceNamePrefix'
          - alb
      SecurityGroups:
        - !GetAtt 'S3GatewaySG.GroupId'
      SubnetMappings:
        - SubnetId: !Ref 'Subnet1'
        - SubnetId: !Ref 'Subnet2'
      Type: application
    Type: AWS::ElasticLoadBalancingV2::LoadBalancer
  ALBHttpListener:
    Properties:
      DefaultActions:
        - ForwardConfig:
            TargetGroups:
              - TargetGroupArn: !Ref 'ALBTargetGroup'
          Type: forward
      LoadBalancerArn: !Ref 'ALB'
      Port: 80
      Protocol: HTTP
    Type: AWS::ElasticLoadBalancingV2::Listener
  ALBTargetGroup:
    Properties:
      Name: !Join
        - '-'
        - - !Ref 'ResourceNamePrefix'
          - log-group
      Port: 80
      Protocol: HTTP
      ProtocolVersion: HTTP1
      TargetGroupAttributes:
        - Key: deregistration_delay.timeout_seconds
          Value: '150'
      TargetType: ip
      VpcId: !Ref 'VpcId'
    Type: AWS::ElasticLoadBalancingV2::TargetGroup
  Cluster:
    Properties:
      ClusterName: !Join
        - '-'
        - - !Ref 'ResourceNamePrefix'
          - cluster
    Type: AWS::ECS::Cluster
  ECSLogGroup:
    Properties:
      LogGroupName: !Join
        - '-'
        - - !Ref 'ResourceNamePrefix'
          - logs
      RetentionInDays: 14
    Type: AWS::Logs::LogGroup
  ECSTaskExecutionPolicy:
    Properties:
      PolicyDocument:
        Statement:
          - Action:
              - logs:CreateLogStream
              - logs:CreateLogGroup
              - logs:PutLogEvents
            Effect: Allow
            Resource:
              - '*'
        Version: '2012-10-17'
      PolicyName: !Join
        - '-'
        - - !Ref 'ResourceNamePrefix'
          - ecs-task-execution-policy
      Roles:
        - !Ref 'ECSTaskExecutionRole'
    Type: AWS::IAM::Policy
  ECSTaskExecutionRole:
    Properties:
      AssumeRolePolicyDocument:
        Statement:
          - Action:
              - sts:AssumeRole
            Effect: Allow
            Principal:
              Service:
                - ecs-tasks.amazonaws.com
        Version: '2012-10-17'
      Description: An IAM role to enable ECS agents to perform AWS operations such as creating CloudWatch logs.
      RoleName: !Join
        - '-'
        - - !Ref 'ResourceNamePrefix'
          - ecs-task-execution-role
    Type: AWS::IAM::Role
  ECSTaskPolicy:
    Properties:
      PolicyDocument:
        Statement:
          - Action:
              - s3:GetObject
              - s3:ListBucket
            Effect: Allow
            Resource:
              - !Sub
                - arn:aws:s3:::${bucketName}/*
                - bucketName: !Ref 'NewBucketName'
              - !Sub
                - arn:aws:s3:::${bucketName}
                - bucketName: !Ref 'NewBucketName'
        Version: '2012-10-17'
      PolicyName: !Join
        - '-'
        - - !Ref 'ResourceNamePrefix'
          - ecs-task-policy
      Roles:
        - !Ref 'ECSTaskRole'
    Type: AWS::IAM::Policy
  ECSTaskRole:
    Properties:
      AssumeRolePolicyDocument:
        Statement:
          - Action:
              - sts:AssumeRole
            Effect: Allow
            Principal:
              Service:
                - ecs-tasks.amazonaws.com
        Version: '2012-10-17'
      Description: An IAM role to enable ECS containers to perform AWS operations such as accessing S3 buckets.
      RoleName: !Join
        - '-'
        - - !Ref 'ResourceNamePrefix'
          - ecs-task-role
    Type: AWS::IAM::Role
  S3Bucket:
    Properties:
      BucketName: !Ref 'NewBucketName'
    Type: AWS::S3::Bucket
  S3GatewaySG:
    Properties:
      GroupDescription: Security group for NGINX S3 Gateway Infra
      GroupName: !Join
        - '-'
        - - !Ref 'ResourceNamePrefix'
          - sg
      SecurityGroupEgress:
        - CidrIp: '0.0.0.0/0'
          Description: Allow all outbound IPv4 traffic
          IpProtocol: '-1'
        - CidrIpv6: ::/0
          Description: Allow all outbound IPv6 traffic
          IpProtocol: '-1'
      SecurityGroupIngress:
        - CidrIp: '0.0.0.0/0'
          Description: Allow HTTP Traffic on 80
          FromPort: 80
          IpProtocol: tcp
          ToPort: 80
        - CidrIp: '0.0.0.0/0'
          Description: Allow HTTPS Traffic on 443
          FromPort: 443
          IpProtocol: tcp
          ToPort: 443
      VpcId: !Ref 'VpcId'
    Type: AWS::EC2::SecurityGroup
  S3GatewayService:
    DependsOn: ALBHttpListener
    Properties:
      Cluster: !Ref 'Cluster'
      DesiredCount: 1
      LaunchType: FARGATE
      LoadBalancers:
        - ContainerName: !Ref 'ContainerName'
          ContainerPort: 80
          TargetGroupArn: !Ref 'ALBTargetGroup'
      NetworkConfiguration:
        AwsvpcConfiguration:
          AssignPublicIp: ENABLED
          SecurityGroups:
            - !GetAtt 'S3GatewaySG.GroupId'
          Subnets:
            - !Ref 'Subnet1'
            - !Ref 'Subnet2'
      ServiceName: !Join
        - '-'
        - - !Ref 'ResourceNamePrefix'
          - service
      TaskDefinition: !Ref 'TaskDefinition'
    Type: AWS::ECS::Service
  TaskDefinition:
    DependsOn: S3Bucket
    Properties:
      ContainerDefinitions:
        - Environment:
            - Name: ALLOW_DIRECTORY_LIST
              Value: 'true'
            - Name: AWS_SIGS_VERSION
              Value: '4'
            - Name: S3_BUCKET_NAME
              Value: !Ref 'S3Bucket'
            - Name: S3_REGION
              Value: !Ref 'AWS::Region'
            - Name: S3_SERVER_PORT
              Value: '443'
            - Name: S3_SERVER_PROTO
              Value: https
            - Name: S3_SERVER
              Value: !Join
                - .
                - - s3
                  - !Ref 'AWS::Region'
                  - amazonaws.com
            - Name: S3_STYLE
              Value: default
            - Name: S3_DEBUG
              Value: 'false'
          Image: ghcr.io/nginxinc/nginx-s3-gateway/nginx-oss-s3-gateway:latest-njs-oss-20220310
          LogConfiguration:
            LogDriver: awslogs
            Options:
              awslogs-group: !Ref 'ECSLogGroup'
              awslogs-region: !Ref 'AWS::Region'
              awslogs-stream-prefix: ecs
          Name: !Ref 'ContainerName'
          PortMappings:
            - ContainerPort: 80
              HostPort: 80
      Cpu: '1024'
      ExecutionRoleArn: !GetAtt 'ECSTaskExecutionRole.Arn'
      Family: !Ref 'ContainerName'
      Memory: '2048'
      NetworkMode: awsvpc
      RequiresCompatibilities:
        - FARGATE
      TaskRoleArn: !GetAtt 'ECSTaskRole.Arn'
    Type: AWS::ECS::TaskDefinition
```
</details>

## Usage

- Update the following 4 parameters in the `Parameters` section of the CloudFormation file for your specific AWS account:
  - `NewBucketName` - any S3 bucket name. Remember that S3 bucket names must be globally unique
  - `VpcId` - Any VPC ID on your AWS account
  - `Subnet1` - Any subnet ID that's in the VPC used above
  - `Subnet2` - Any subnet ID that's in the VPC used above
- Run the following command to deploy the stack (this assumes you have the AWS CLI & credentials setup correctly on your host machine):

  ```sh
  aws cloudformation create-stack \
    --stack-name nginx-s3-gateway \
    --capabilities CAPABILITY_NAMED_IAM \
    --template-body file://cf.yaml
  ```

- Wait for the CloudFormation Stack deployment to complete (can take about 3-5 minutes)
  - You can query the stack status with this command:
    ```sh
    aws cloudformation describe-stacks \
      --stack-name nginx-s3-gateway \
      --query "Stacks[0].StackStatus"
    ```
- Wait until the query above shows `"CREATE_COMPLETE"`
- Run the following command to get the URL used to access the service:
  ```sh
  aws cloudformation describe-stacks \
    --stack-name nginx-s3-gateway \
    --query "Stacks[0].Outputs[0].OutputValue"
  ```
- View the container logs in CloudWatch from the AWS web console
- Run the following command to delete the stack and all resources:
  ```sh
  aws cloudformation delete-stack \
    --stack-name nginx-s3-gateway
  ```

dekobon:
@FlorianSW 

> * Lastly, we will not be able to merge until the next version of njs is released that fixes a bug that we hit with the instance profile credential code path.
> What is this about?

When an exception happens within a promise (and even if it is caught), it will result in a segmentation fault of njs. It is fixed in the following two commits to njs:
https://hg.nginx.org/njs/rev/bede4b8a693a
https://hg.nginx.org/njs/rev/040d1b318a15

We rely on catching an exception in order to not have a race condition here:  https://github.com/nginxinc/nginx-s3-gateway/blob/issue/8/common/etc/nginx/include/s3gateway.js#L257

dekobon:
@ajschmidt8 I added the docs you contributed above to the branch under your name. I hope that is ok.

Commit: https://github.com/nginxinc/nginx-s3-gateway/commit/20a156e985d027051ca1933de1a4497c9e01b42c

ajschmidt8:
> @ajschmidt8 I added the docs you contributed above to the branch under your name. I hope that is ok.
> 
> Commit: [20a156e](https://github.com/nginxinc/nginx-s3-gateway/commit/20a156e985d027051ca1933de1a4497c9e01b42c)

Absolutely. No problems here. It's a little unrefined in its current state since I was just re-using some stuff I had already written. I will try to tweak it later this week to spin up a VPC with subnets too so that those variables don't need to be changed per user.

ajschmidt8:
Any updates here? What's left on the to-do list?

dekobon:
@ajschmidt8 - Everything is fully functioning to my knowledge in the [issue/8 branch](https://github.com/nginxinc/nginx-s3-gateway/tree/issue/8). This branch is publishing a Docker image to the github container registry, so it can be used as-is right now.

I'm waiting for a new njs release before merging into master in order for a bug affecting instance profile support to be resolved.

cyuste:
Hello,
Great work! Looking forward to be merged in master.
Just one question: is this branch currently ready to test, I tried to run it in docker with ec2 credentials and got the following error
```
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/default.conf.template to /etc/nginx/conf.d/default.conf
20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/gateway/v2_headers.conf.template to /etc/nginx/conf.d/gateway/v2_headers.conf
20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/gateway/v2_js_vars.conf.template to /etc/nginx/conf.d/gateway/v2_js_vars.conf
20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/gateway/v4_headers.conf.template to /etc/nginx/conf.d/gateway/v4_headers.conf
20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/gateway/v4_js_vars.conf.template to /etc/nginx/conf.d/gateway/v4_js_vars.conf
20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/upstreams.conf.template to /etc/nginx/conf.d/upstreams.conf
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2022/05/12 12:33:54 [emerg] 1#1: SyntaxError: "signedHeaders" has already been declared in s3gateway.js:494
nginx: [emerg] SyntaxError: "signedHeaders" has already been declared in s3gateway.js:494
```
I didn't really investigated the issue, I wanted to clarify its state before going crazy :)

dekobon:
Hi @cyuste the gateway configuration should be working in its current state. The only reason that I've held off from merging is that I'm waiting for the latest release of njs to drop for nginx plus. Can you share what environment variables you started the gateway with (please redact sensitive data)?

Also, can you share how you pulled in the container image?

ajschmidt8:
> The only reason that I've held off from merging is that I'm waiting for the latest release of njs to drop for nginx plus

@dekobon, do you know when that will be? Just trying to gauge when this will be merged into the main branch.

cyuste:
Hello @dekobon ,

This is what I did:

Download repo from https://github.com/FlorianSW/nginx-s3-gateway/archive/refs/heads/issue/8.zip
Build:  docker build -f Dockerfile.oss -t nginx-oss-s3-gateway:issue-8  .

Create settings file
```
S3_BUCKET_NAME=redacted-bucket-name
S3_SERVER=s3-eu-west-1.amazonaws.com
S3_SERVER_PORT=443
S3_SERVER_PROTO=https
S3_REGION=eu-west-1
S3_STYLE=virtual
S3_DEBUG=false
AWS_SIGS_VERSION=4
ALLOW_DIRECTORY_LIST=false
```
Run: docker run --env-file ./settings -p8433:80 --name nginx-oss-s3-gateway nginx-oss-s3-gateway:issue-8  

The error is the same from the previous comment. 
Thank you for your help

dekobon:
Hi @cyuste , I think I may have found the issue. In the steps you provided above, you are pulling code from @FlorianSW 's branch. I've taken Florian's code and made additional changes. The latest code is in the issue/8 branch of this repository. As such, the download url to use above should be: https://github.com/nginxinc/nginx-s3-gateway/archive/refs/heads/issue/8.zip

FlorianSW:
I should probably remove my branch there, so there is no confusion 🙈

cyuste:
Thank you both very much for your support, works perfectly, I would run now some tests to check if it fits our needs. 

Great work!

dekobon:
This PR is merged in commit: 0df77d631ea6873468d9c862187d7d524db0b7b4

PR #32, opened 2022-05-13
Expose proxy_cache_valid as configurable environment variable.

goroi:
There are use cases where you want to set different than defaults proxy_cache_valid values.
Exposing these values make the use of this image more flexible.


dekobon:
This branch will be merged into master next week: https://github.com/nginxinc/nginx-s3-gateway/tree/issue/8

I will work on merging these changes with the newer code next week. If you want to get a head start, please consider modifying this PR to work against that branch.

In terms of the code changes, they look good, but the documentation will also need to be updated to reflect the additional environment variables.

My apologies for any extra work that this may have caused.

Issue #38, opened 2022-05-24
Error 404 Handling During Debugging


rmcsqrd:
### Discussed in https://github.com/nginxinc/nginx-s3-gateway/discussions/37

<div type='discussions-op-text'>

<sup>Originally posted by **rmcsqrd** May 24, 2022</sup>
Hi there, thanks for making this - I got it working successfully and it is very useful.

I ran into issues with the REST authentication because MFA was enabled for the AWS account tied to my access key/secret key pair and ended up needing to make a new role with reduced permissions and no MFA (obviously this is not at all related to your package).

During the debugging process I found the the default 404 response behavior was a bit of a red herring in the debugging process (but understand its utility for production systems). I spent a lot of time trying a lot of different things before I dug into the actual nginx conf file and found the part of the configuration that [changed the error response](https://github.com/nginxinc/nginx-s3-gateway/blob/master/common/etc/nginx/templates/default.conf.template#L107-L112). Once I commented these lines out locally I saw I was actually getting a 403 Access Denied error which helped narrow down the debugging process. 

I think it would be useful to mention this within the "Getting Started Guide" as a potential way to debug any authentication error. I also think it might be useful to mention the MFA thing as well. I am happy to file a documentation PR calling this out if you think it would be useful. </div>

rmcsqrd:
@dekobon Moved this to an issue for tracking, I'll submit a PR soon

dekobon:
Closed in PR #39

PR #39, opened 2022-05-24
update getting_started w/ debugging information

rmcsqrd:
In response to #38. Also the "contents" section wasn't formatting on different lines when rendered so added spaces after each line to correct that.


dekobon (approved):
LGTM

PR #43, opened 2022-08-01
Fix auth v4 for directory listing

iNecas:
I was getting `SignatureDoesNotMatch` errors when using directory
listing with v4 authentication. Seems like the problem was
the requests were sent with trailing `/` while the signature
was calculated without it.

It doesn't seem we need the trailing space there (I have checked
only with Ceph S3 API, so more testing with AWS welcome), but seems
like the code should be broken for every back-end, unless I'm missing
something.

### DIFF
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -348,7 +348,7 @@ function s3uri(r) {
     if (allow_listing) {
         var queryParams = _s3DirQueryParams(uriPath, r.method);
         if (queryParams.length > 0) {
-            path = basePath + '/?' + queryParams;
+            path = basePath + '?' + queryParams;
         } else {
             path = basePath + uriPath;
         }


dekobon:
Hi @iNecas, I was unable to reproduce the error this when running on AWS or Minio. That said, there are a lot of fiddly differences between the different S3 compatible services out there. This may be one of those cases. Perhaps AWS is more permissive in what is sent.

dekobon (approved):
Great contribution! :+1:

dekobon:
Based on my tests, applying this change seems to work with AWS and Minio and does not have a negative impact. I'm going to go ahead and merge.

iNecas:
Thanks for fast response. Cheers!

dekobon:
This fix was lost in a bad merge. It has been re-applied in commit: f48542933adc3853e74d5fb162d0014be417c90a

PR #47, opened 2022-08-10
Support unit test on Windows machine

Gamecock:
Addresses issues running unt test on windows except unicode characters in URLs.

Primary changes:
1)  adding quotes to variable expansions:  `"${variable}"` vs `${variable}` to handle spaces in path names.
2) MSYS_NO_PATHCONV=1 added to resolve automatic path conversion  https://github.com/docker/for-win/issues/6754#issuecomment-629702199

Based on #42 will be cleaner after it is merged.

@dekobon  or @victrid can you confirm no impact on linux unit test?


dekobon (approved):
LGTM!

PR #51, opened 2022-08-22
[idea] Add support for authorisation check on each request

dotZoki:
This PR is more of an idea/suggestion to have option to add auth checking on app API before serving an image from s3.
I don't think it is good to merge but rather maybe a starting point for a feature. Feel free to close PR as soon/if this is noted somewhere to be implemented.


dekobon:
This type of "feature" is just configuration of nginx rather than changing the S3 authentication workflow. It would make for a good "example", that could live under the [examples/](/examples) directory. I would use the main s3 gateway as the parent container and then write a Dockerfile that would extend it and layer in the configuration that you want to build.

I would ask that you include a detailed README that explains:
* The use case
* The reason why you are taking this approach
* How things work / why things were changed from their default

dotZoki:
OK, thanks for input. I'll open new PR when I prepare things.

PR #55, opened 2022-09-28
Support for regional STS endpoint detection

djnalluri:
The appropriate STS endpoint gets detected via the `AWS_STS_REGIONAL_ENDPOINTS` and `AWS_REGION` env variables.

This change is untested as I don't have an environment available for testing currently.

### DIFF
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -1048,7 +1048,7 @@ async function _fetchWebIdentityCredentials(r) {
         if (sts_regional === 'regional') {
             var region = process.env['AWS_REGION'];
             if (region) {
-                sts_endpoint = 'https://sts.' + region + '.amazonaws.com';
+                sts_endpoint = `https://sts.${region}.amazonaws.com`;
             } else {
                 throw 'Missing required AWS_REGION env variable';
             }
@@ -1059,7 +1059,7 @@ async function _fetchWebIdentityCredentials(r) {
 
     var token = fs.readFileSync(process.env['AWS_WEB_IDENTITY_TOKEN_FILE']);
     
-    var params = "Version=2011-06-15&Action=AssumeRoleWithWebIdentity&RoleArn=" + arn + "&RoleSessionName=" + name + "&WebIdentityToken=" + token;
+    var params = `Version=2011-06-15&Action=AssumeRoleWithWebIdentity&RoleArn=${arn}&RoleSessionName=${name}&WebIdentityToken=${token}`;
 
     var response = await ngx.fetch(sts_endpoint + "?" + params, {
         headers: {
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -1044,8 +1044,16 @@ async function _fetchWebIdentityCredentials(r) {
 
     var sts_endpoint = process.env['STS_ENDPOINT'];
     if (!sts_endpoint) {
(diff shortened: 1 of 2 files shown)


common/etc/nginx/include/s3gateway.js
```diff
@@ -1041,7 +1041,22 @@ async function _fetchEC2RoleCredentials() {
...
     var name  = process.env['HOSTNAME'] || 'nginx-s3-gateway';
-    var sts_endpoint = process.env['STS_ENDPOINT'] || 'https://sts.amazonaws.com';
+
+    var sts_endpoint = process.env['STS_ENDPOINT'];
+    if (!sts_endpoint) {
+        var sts_regional = process.env['AWS_STS_REGIONAL_ENDPOINTS'] || 'global';
+        if (sts_regional === 'regional') {
+            var region = process.env['AWS_REGION'];
+            if (region) {
+                sts_endpoint = 'https://sts.' + region + '.amazonaws.com';
```
dekobon, line 1051: Njs now supports interpolation, so you can do:
```
`https://sts.${region}.amazonaws.com`
```

common/etc/nginx/include/s3gateway.js
```diff
@@ -1041,7 +1041,22 @@ async function _fetchEC2RoleCredentials() {
...
+                throw 'Missing required AWS_REGION env variable';
+            }
+        } else {
+            sts_endpoint = 'https://sts.amazonaws.com';
+        }
+    }
+
     var token = fs.readFileSync(process.env['AWS_WEB_IDENTITY_TOKEN_FILE']);
     
     var params = "Version=2011-06-15&Action=AssumeRoleWithWebIdentity&RoleArn=" + arn + "&RoleSessionName=" + name + "&WebIdentityToken=" + token;
```
dekobon, line 1062: Same as above, this string can be composed using interpolation.

dekobon:
@djnalluri I may be able to get someone with an environment who can test. However, I'm unsure how to test. Can you write instructions for testing the feature as well as update the documentation to reflect the newly added environment variables?

djnalluri:
@dekobon I just updated the documentation and shifted some things around in there. It now includes a YAML to minimally create a deployment (untested as well).  We would need to test 2 different configurations to ensure this is working. One test with the `eks.amazonaws.com/sts-regional-endpoints` annotation set to `"true"` and another with it set to false. We may also need to inject a log statement to ensure the correct endpoint is getting picked in both cases.

common/etc/nginx/include/s3gateway.js
```diff
@@ -1041,10 +1041,25 @@ async function _fetchEC2RoleCredentials() {
 async function _fetchWebIdentityCredentials(r) {
     var arn   = process.env['AWS_ROLE_ARN'];
     var name  = process.env['HOSTNAME'] || 'nginx-s3-gateway';
-    var sts_endpoint = process.env['STS_ENDPOINT'] || 'https://sts.amazonaws.com';
+
+    var sts_endpoint = process.env['STS_ENDPOINT'];
+    if (!sts_endpoint) {
+        var sts_regional = process.env['AWS_STS_REGIONAL_ENDPOINTS'] || 'global';
```
dekobon, line 1047: Could you please add docs for the `AWS_STS_REGIONAL_ENDPOINTS` environment variable as well?

dekobon: On second thought, there is no need to delaying merging on account of docs.

PR #56, opened 2022-10-04
Fix credentials caching

dm3ch:
I found two issues while I was trying to figure out what's wrong with credentials caching:
- Date object was constructed with Unix timestamp value, while it was assuming that it would be milliseconds timestamp, so i added `* 1000`
- Based on comment it seems that `maxValidityOffsetMs` should be 4.5 minutes, but it was 0.45 minute

### DIFF
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -882,7 +882,7 @@ function _require_env_var(envVarName) {
  *
  * @type {number}
  */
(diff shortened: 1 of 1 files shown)


dekobon:
This is great work! Thank you for tracking down the bug and doing the fix. The PR passes all of the tests with OSS/Plus and the latest versions of njs. I did a quick smoke test on AWS seeing if everything worked after waiting for 5 minutes and it had no problems.

Issue #53, opened 2022-08-31
Filenames with brackets () return 404 errors


sixem:
Hi,

I'm running the the stand-alone container on one of my servers. Problem is, whenever one of the files stored in the S3 container contains brackets, the gateway will return a 404 error.

An example:

Both `image.jpg` and `image(1).jpg` exists in the same bucket in the same folder.

This:
```bash
curl -I "http://127.0.0.1:8011/image.jpg"
```
Returns:
```
HTTP/1.1 200 OK
Server: nginx
Date: Wed, 31 Aug 2022 17:45:21 GMT
Content-Type: image/jpeg
Content-Length: 466483
Connection: keep-alive
Last-Modified: Wed, 31 Aug 2022 17:44:50 GMT
ETag: "b34cf0ebd821a9f892b361e6812571eb"
Accept-Ranges: bytes
```
While this:
```bash
curl -I "http://127.0.0.1:8011/image(1).jpg"
```
Returns:
```
HTTP/1.1 404 Not Found
Server: nginx
Date: Wed, 31 Aug 2022 17:45:27 GMT
Content-Type: text/html
Content-Length: 146
Connection: keep-alive
```

The gateway log shows this:
```
127.0.0.1 - - [31/Aug/2022:17:45:21 +0000] "HEAD /image.jpg HTTP/1.1" 200 0 "-" "curl/7.64.0" "-"
2022/08/31 17:45:21 [info] 61#61: *159 client 127.0.0.1 closed keepalive connection
127.0.0.1 - - [31/Aug/2022:17:45:27 +0000] "HEAD /image(1).jpg HTTP/1.1" 404 0 "-" "curl/7.64.0" "-"
2022/08/31 17:45:27 [info] 61#61: *161 client 127.0.0.1 closed keepalive connection
```

Server details:
```
Debian 8.3.0-6 GNU/Linux 10 (buster)
Linux version 4.19.0-14-cloud-amd64
```

dekobon:
Hi @sixem ,

What S3 service are you using on the backend? Is it AWS S3 or something else?

sixem:
> Hi @sixem ,
> 
> What S3 service are you using on the backend? Is it AWS S3 or something else?

I'm using AWS S3.

cpeled:
Hi,
it seems that the same happens for every escaped uri path.
the s3 url is generated correctly (i compared it with aws s3 object url) but probably something is wrong with the canonical v4 request.
[...]

PR #57, opened 2022-10-06
Fix 404 errors for filenames containing certain characters

sixem:
Hi,

I was experiencing 404 errors when using the gateway with filenames which mainly contained brackets `()` (#53). This pull request attempts to fix that by applying additional character encodings to the native `encodeURIComponent` function that is being called during SigV4 creation when the URI path is being escaped.

I've added tests which have passed for me successfully. I've also tested the built docker image with AWS S3, and everything appears to work fine for me there as well.

I'm not too familiar with the process of handling S3 requests, so any input, changes or suggestions are very much welcome.

Summed up, this PR attempts to fix the 404 errors encountered when filenames contain any of these characters: `()'!*`.

### DIFF
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -358,7 +358,7 @@ function s3uri(r) {
         }
     } else {
         if (provide_index_page  && _isDirectory(uriPath) ) {
-               uriPath += INDEX_PAGE;
+            uriPath += INDEX_PAGE;
         }
         path = basePath + uriPath;
     }
@@ -571,6 +571,7 @@ function _buildSignatureV4(r, amzDatetime, eightDigitDate, creds, bucket, region
     } else {
         uri = _escapeURIPath(s3uri(r));
     }
+
     var canonicalRequest = _buildCanonicalRequest(method, uri, queryParams, host, amzDatetime, creds.sessionToken);
 
     _debug_log(r, 'AWS v4 Auth Canonical Request: [' + canonicalRequest + ']');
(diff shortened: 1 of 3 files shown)


dekobon:
Thank you for your contribution! This makes fixing the issue easier. I'd like to iterate on your PR with a few changes. In order to do so, I created a new branch: https://github.com/nginxinc/nginx-s3-gateway/tree/issue/53

Once I have made my changes I will do a PR from that branch, copy you in, and ask you for a review.

sixem:
Sounds good!

Just so that you are aware of it, I should also mention that filenames with plus symbols (`+`) also return 404 errors. As also discovered by @kemoT1990TM in the issue, this method does not solve that error, which makes sense because `encodeURIComponent()` should encode these symbols by default anyways.

I was messing around, and through debugging I did find a way to solve the plus symbol errors for AWS S3, but the tests on MinIO failed, so I assume there are some discrepancies there, and I'm unsure of how to go forward with that issue unfortunately.

dekobon:
Let's resolve it on the PR for the branch together.

kemoT1990TM:
> Sounds good!
> 
> Just so that you are aware of it, I should also mention that filenames with plus symbols (`+`) also return 404 errors. As also discovered by @kemoT1990TM in the issue, this method does not solve that error, which makes sense because `encodeURIComponent()` should encode these symbols by default anyways.
> 
> I was messing around, and through debugging I did find a way to solve the plus symbol errors for AWS S3, but the tests on MinIO failed, so I assume there are some discrepancies there, and I'm unsure of how to go forward with that issue unfortunately.

Hello @sixem are You able to give me a tip how to resolve issue with + sign? I am trying to resolve it but still no luck. We are using S3 compatible storage provided by our clod provider.

dekobon:
Closing this PR in favor of this PR: https://github.com/nginxinc/nginx-s3-gateway/pull/58

PR #58, opened 2022-10-10
Resolve #53 by fixing multiple encoding problems

### LINKED ISSUE #53: Filenames with brackets () return 404 errors

dekobon:
Fixes #53 

This PR addresses problems with:
 * Parenthesis in URLs
 * Some Unicode in URLs (like bidi Hebrew characters)
 * Plus signs in URLs

This PR also adds unit tests in order to catch future encoding issues.

### DIFF
--- a/common/etc/nginx/include/listing.xsl
+++ b/common/etc/nginx/include/listing.xsl
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
-<xsl:stylesheet version="1.1" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+<xsl:stylesheet version="1.1" xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:str="http://exslt.org/strings" extension-element-prefixes="str">
     <xsl:output method="html" encoding="utf-8" indent="yes"/>
+    <xsl:strip-space elements="*" />
 
     <xsl:template match="/">
         <xsl:choose>
@@ -81,7 +82,7 @@
                           select="substring-after(text(), $globalPrefix)"/>
             <tr>
                 <td>
-                    <a href="/{text()}">
+                    <a><xsl:attribute name="href">/<xsl:call-template name="encode-uri"><xsl:with-param name="uri" select="text()"/></xsl:call-template>/</xsl:attribute>
                         <xsl:value-of select="$dirName"/>
                     </a>
                 </td>
@@ -103,7 +104,8 @@
             <xsl:variable name="size" select="*[local-name()='Size']/text()"/>
             <tr>
                 <td>
-                    <a href="/{$key}">
+                    <a>
+                        <xsl:attribute name="href">/<xsl:call-template name="encode-uri"><xsl:with-param name="uri" select="$key"/></xsl:call-template></xsl:attribute>
                         <xsl:value-of select="$fileName"/>
                     </a>
                 </td>
@@ -115,5 +117,22 @@
                 </td>
             </tr>
         </xsl:if>
+    </xsl:template>
+    <!-- This template escapes the URI such that symbols or unicode characters are
+         encoded so that they form a valid link that NGINX can parse -->
+    <xsl:template name="encode-uri">
+        <xsl:param name="uri"/>
+        <xsl:for-each select="str:split($uri, '/')">
+            <xsl:variable name="encoded" select="str:encode-uri(., 'true', 'UTF-8')" />
+            <xsl:variable name="more-encoded" select="
+                str:replace(
+                    str:replace(
+                        str:replace(
+                            str:replace(
+                                str:replace($encoded, '@', '%40'), '(', '%28'),
+                        ')', '%29'),
+                    '!', '%21'),
+                '*', '%2A')" />
+            <xsl:value-of select="$more-encoded" /><xsl:if test="position() != last()">/</xsl:if></xsl:for-each>
     </xsl:template>
 </xsl:stylesheet>
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -354,13 +354,14 @@ function s3uri(r) {
         if (queryParams.length > 0) {
             path = basePath + '?' + queryParams;
         } else {
-            path = basePath + uriPath;
+            path = _escapeURIPath(basePath + uriPath);
         }
     } else {
+        // This is a path that will resolve to an index page
         if (provide_index_page  && _isDirectory(uriPath) ) {
             uriPath += INDEX_PAGE;
         }
-        path = basePath + uriPath;
+        path = _escapeURIPath(basePath + uriPath);
     }
 
     _debug_log(r, 'S3 Request URI: ' + r.method + ' ' + path);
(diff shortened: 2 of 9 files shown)


dekobon:
This is the new PR for this issue (#53). Going forward, let's make changes against this branch and then submit additional PRs.

@sixem @kemoT1990TM @cpeled

dekobon:
I believe this PR now solves the problems with symbols and Unicode characters in object names. However, this PR does not yet have a fix for the index page generating incorrect links to the objects. For example, the index page will generate a URL like the following `http://localhost/a/plus+plus.txt` when it should in fact be `http://localhost/a/plus%2Bplus.txt`.

As long as the URL in the HTML `a href` is not properly encoded, the object may not be consistently returned. In some cases, NGINX makes a best guess as to what it should be, but in many cases it does.

This leaves us with a difficult problem. The index page is generated using XSLT based on the response of an unescaped directory listing returned by S3. There seems to be no support to XSL extensions for the index page. I see a few non-exclusive options:

1. Call this PR good enough, merge it, and then open a new issue for index page links.
2. Fix link encoding in the index page via Javascript.
3. Try to search for another solution.

Thoughts?

dekobon:
@kemoT1990TM What happens when you try to request an object with the plus sign encoded as `%2B`?

kemoT1990TM:
> @kemoT1990TM What happens when you try to request an object with the plus sign encoded as `%2B`?
In this case the error message is:
S3 error response: RequestId=1665046259594210, TraceId=1f8ae499d25b4a36, Resource=/bucket-dev/Product/1.4.0/en/filetype-eppRequisitionType+register-eppHistory.html, HTTP Method=GET, HTTP Status Code=403, X-Forwarded-For: 'xx.xx.xx.xx', ErrorMsg=SignatureDoesNotMatch, ErrorType=ClientAuthorization

dekobon:
> > @kemoT1990TM What happens when you try to request an object with the plus sign encoded as `%2B`?
> > In this case the error message is:
> > S3 error response: RequestId=1665046259594210, TraceId=1f8ae499d25b4a36, Resource=/bucket-dev/Product/1.4.0/en/filetype-eppRequisitionType+register-eppHistory.html, HTTP Method=GET, HTTP Status Code=403, X-Forwarded-For: 'xx.xx.xx.xx', ErrorMsg=SignatureDoesNotMatch, ErrorType=ClientAuthorization

To be clear you are still getting this error when using this (issue/53) branch?

sixem (commented):
I would suggest maybe opening another PR to fix the indexing issues with plus symbols. After doing some more testing, it appears to me that using `%2B` instead of `+` has always worked prior to this fix, which makes sense because `A-Z a-z 0-9 - _ . ! ~ * ' ( )` are the only characters not being encoded by default when using `encodeURIComponent()`, and this PR does address the issue with these unescaped characters in particular, so the indexing issue is unrelated to this PR, I think.

I must've gotten thrown off by either browser or web server caching when during testing and switching between `+` and `%2B` ended up always returning 404 errors. Testing it now with only `%2B`, it appears to work for me at least. If anyone else could also confirm this that would be great.

I'm unsure why @kemoT1990TM is still experiencing these issues even with `%2B`, so we may have to look further into that.

dekobon:
> I would suggest maybe opening another PR to fix the indexing issues with plus symbols. After doing some more testing, it appears to me that using `%2B` instead of `+` has always worked prior to this fix, which makes sense because `A-Z a-z 0-9 - _ . ! ~ * ' ( )` are the only characters not being encoded by default when using `encodeURIComponent()`, and this PR does address the issue with these unescaped characters in particular, so the indexing issue is unrelated to this PR, I think.
> 
> I must've gotten thrown off by either browser or web server caching when during testing and switching between `+` and `%2B` ended up always returning 404 errors. Testing it now with only `%2B`, it appears to work for me at least. If anyone else could also confirm this that would be great.
> 
> I'm unsure why @kemoT1990TM is still experiencing these issues even with `%2B`, so we may have to look further into that.

I believe that I've fixed the lingering issues with the latest commit. @kemoT1990TM and @sixem , if you have some time could you verify? So far things seem to be working as I connect directly to AWS and when I test against MinIO.

sixem:
> > I would suggest maybe opening another PR to fix the indexing issues with plus symbols. After doing some more testing, it appears to me that using `%2B` instead of `+` has always worked prior to this fix, which makes sense because `A-Z a-z 0-9 - _ . ! ~ * ' ( )` are the only characters not being encoded by default when using `encodeURIComponent()`, and this PR does address the issue with these unescaped characters in particular, so the indexing issue is unrelated to this PR, I think.
> > I must've gotten thrown off by either browser or web server caching when during testing and switching between `+` and `%2B` ended up always returning 404 errors. Testing it now with only `%2B`, it appears to work for me at least. If anyone else could also confirm this that would be great.
> > I'm unsure why @kemoT1990TM is still experiencing these issues even with `%2B`, so we may have to look further into that.
> 
> I believe that I've fixed the lingering issues with the latest commit. @kemoT1990TM and @sixem , if you have some time could you verify? So far things seem to be working as I connect directly to AWS and when I test against MinIO.

I can confirm that filenames using either `+` or `%2B` work for me now. Tested on both AWS S3 and Backblaze. :+1:

dekobon:
In the last commit, somehow I was able to remember enough XSLT to fix the directory listing issue. I'm quite happy with this PR so far. I really needed the push from @sixem to get started. Thank you and @kemoT1990TM !

I'm going to wait a day before merging to give all of you a chance to test.

cpeled:
Hi.
for some reason, '+' does not work for me with AWS S3. i keep getting 404.
i upload a filename "n ginx.conf" (with space)
this is  the url aws attach the object:  https://<BUCKET_NAME>.eu-central-1.amazonaws.com/n+ginx.conf
when building the s3uri in the gw:  /n%2Bginx.conf

and it just returns 404

any idea?

Issue #62, opened 2022-10-13
Possible to support OPTIONS requests for CORS?


worace:
**Is your feature request related to a problem? Please describe.**

Just tried this project out for the first time and it's great. Docs were clear and easy to get started.

One issue I ran into is that it appears OPTIONS requests currently are rejected by the JS module: https://github.com/nginxinc/nginx-s3-gateway/blob/master/common/etc/nginx/include/s3gateway.js#L410-L415

It would be great to be able to enable CORS settings since that would make it easier to use this proxy directly from the browser for client-side applications.

Is it possible to add support there, along with some means to configure CORS settings? 

**Describe the solution you'd like**

**Describe alternatives you've considered**

I tried adding CORS settings via an nginx conf file under `/etc/nginx/conf.d/` but I think the JS module is taking precedence and rejecting these requests before my settings take effect.

That said, it's definitely possible I am configuring something wrong...I am by no means an expert on either NGINX or CORS.

Here is the dockerfile I was using:

```
FROM ghcr.io/nginxinc/nginx-s3-gateway/nginx-oss-s3-gateway:latest

COPY nginx.conf /etc/nginx/conf.d/default.conf
```

And this is the config file i tried:

```
server {
    echo
    listen       80;
    server_name  localhost;

    #access_log  /var/log/nginx/host.access.log  main;

    location / {
        echo "hello"
        if ($request_method = 'OPTIONS') {
           echo "method is options"
           add_header 'Access-Control-Allow-Origin' '*';
           add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
           #
           # Custom headers and headers various browsers *should* be OK with but aren't
           #
           add_header 'Access-Control-Allow-Headers' 'DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range';
           #
           # Tell client that this pre-flight info is valid for 20 days
           #
           add_header 'Access-Control-Max-Age' 1728000;
           add_header 'Content-Type' 'text/plain; charset=utf-8';
           add_header 'Content-Length' 0;
           return 204;
        }
        if ($request_method = 'POST') {
            add_header 'Access-Control-Allow-Origin' '*' always;
            add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS' always;
[...]

PR #63, opened 2022-10-14
Add CORS headers, configurable with a CORS_ALLOW_ALL setting

worace:


worace:
https://github.com/nginxinc/nginx-s3-gateway/issues/62

worace:
Example before/after:

```
$ curl -I -X OPTIONS http://localhost/user/horace/wof_neighborhoods.fgb
HTTP/1.1 405 Not Allowed
Server: nginx
Date: Fri, 14 Oct 2022 21:27:09 GMT
Content-Type: text/html
Content-Length: 150
Connection: keep-alive
Allow: GET, HEAD

$ curl -I -X OPTIONS http://localhost/user/horace/wof_neighborhoods.fgb
HTTP/1.1 204 No Content
Server: nginx
Date: Fri, 14 Oct 2022 21:27:23 GMT
Connection: keep-alive
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET, POST, OPTIONS
Access-Control-Allow-Headers: DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range
Access-Control-Max-Age: 1728000
Content-Length: 0
```

dekobon:
Thank you for the PR! I see it is still a draft, so you must be still iterating. All in all this looks, good. I do wonder if it is possible to add an integration test that makes sure that CORS is working as per expectation when enabled.

dav-pascual:
Is there any progress on this review? @dekobon 
This would be a very nice feature to have :)

dekobon:
Hi @dav-pascual and @worace my apologies for the late response. After a lot of thinking, I'd like to implement this in the nginx configuration rather than co-locate it in the AWS/S3 specific JS file.

I've created this [PR](https://github.com/nginxinc/nginx-s3-gateway/pull/74) and I would like your review and testing on it if possible.

worace:
@dekobon I think that sounds like a great approach; probably more similar to what other users with prior familiarity with nginx would expect

PR #71, opened 2022-11-18
Fix startup when using IAM roles for service accounts

### LINKED ISSUE #70: Startup fails when using IAM Roles for Service Accounts
nginx-s3-gateway container does not start when using IAM roles for service accounts (in EKS) as authentication mechanism.

**To Reproduce**
start the container with the following variables:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: example
  template:
    metadata:
      labels:
        app.kubernetes.io/name: example
    spec:
      serviceAccountName: my-sa
      containers:
        - name: nginx-s3-gateway
          image: nginxinc/nginx-s3-gateway:latest
          env:
            - name: AWS_SIGS_VERSION
              value: "4"
            - name: S3_BUCKET_NAME
              value: my-bucket
            - name: S3_REGION
              value: us-east-1
            - name: S3_SERVER
              value: s3.us-east-1.amazonaws.com
            - name: S3_SERVER_PORT
              value: "443"
            - name: S3_SERVER_PROTO
              value: "https"
            - name: S3_STYLE
              value: default
```

causes this error on startup:

```
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/00-check-for-required-env.sh
Required S3_ACCESS_KEY_ID environment variable missing
Required S3_SECRET_KEY environment variable missing
```

**Expected behavior**
The container startup should not fail.

**Your environment**
* Version of the repo: latest (30529460188d4ffbd2ab8d298ce04407258c6ffb)
* Version of the container used: latest (42aa78b1f5d7c316ea93611ccae44170519b434e96070f17b388b07bf4cf4c67)
* S3 backend implementation: AWS
* How you are deploying: EKS
* NGINX type: OSS
* Authentication method: IAM role for service account

jnodorp-jaconi:
Closes #70

### DIFF
--- a/common/docker-entrypoint.d/00-check-for-required-env.sh
+++ b/common/docker-entrypoint.d/00-check-for-required-env.sh
@@ -25,16 +25,34 @@ failed=0
 required=("S3_BUCKET_NAME" "S3_SERVER" "S3_SERVER_PORT" "S3_SERVER_PROTO"
 "S3_REGION" "S3_STYLE" "ALLOW_DIRECTORY_LIST" "AWS_SIGS_VERSION")
 
-if [ ! -z ${AWS_CONTAINER_CREDENTIALS_RELATIVE_URI+x} ]; then
+# Require some form of authentication to be configured.
+
+# a) Using container credentials. This is indicated by AWS_CONTAINER_CREDENTIALS_RELATIVE_URI being set.
+#    See https://docs.aws.amazon.com/sdkref/latest/guide/feature-container-credentials.html
+#    Example: We are running inside an ECS task.
+if [[ -v AWS_CONTAINER_CREDENTIALS_RELATIVE_URI ]]; then
   echo "Running inside an ECS task, using container credentials"
+
+# b) Using Instance Metadata Service (IMDS) credentials, if IMDS is present at http://169.254.169.254.
+#    See https://docs.aws.amazon.com/sdkref/latest/guide/feature-imds-credentials.html.
+#    Example: We are running inside an EC2 instance.
 elif curl --output /dev/null --silent --head --fail --connect-timeout 2 "http://169.254.169.254"; then
   echo "Running inside an EC2 instance, using IMDS for credentials"
+
+# c) Using assume role credentials. This is indicated by AWS_WEB_IDENTITY_TOKEN_FILE being set.
+#    See https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html.
+#    Example: We are running inside an EKS cluster with IAM roles for service accounts enabled.
+elif [[ -v AWS_WEB_IDENTITY_TOKEN_FILE ]]; then
+  echo "Running inside EKS with IAM roles for service accounts"
+
+# If none of the options above is used, require static credentials.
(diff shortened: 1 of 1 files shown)


common/docker-entrypoint.d/00-check-for-required-env.sh
```diff
@@ -29,6 +29,8 @@ if [ ! -z ${AWS_CONTAINER_CREDENTIALS_RELATIVE_URI+x} ]; then
   echo "Running inside an ECS task, using container credentials"
 elif curl --output /dev/null --silent --head --fail --connect-timeout 2 "http://169.254.169.254"; then
   echo "Running inside an EC2 instance, using IMDS for credentials"
+elif [ ! -z ${AWS_WEB_IDENTITY_TOKEN_FILE+x} ]; then
```
dekobon, line 32: Can you add a comment above this line that explains the logic? If I don't know anything, I'll that I see is that if the `AWS_WEB_IDENTITY_TOKEN_FILE` environment variable exists, then we output a string. Please explain what is the significance of doing nothing.

Also, I believe you can do this and it is the same, so you don't need to do double negation. However, please test to make sure I'm not crazy.
```
elif [ -n ${AWS_WEB_IDENTITY_TOKEN_FILE+x} ]; then
```

dekobon (changes requested):
Thank you for the submission! I'm sorry that I was not able to respond sooner.

common/docker-entrypoint.d/00-check-for-required-env.sh
jnodorp-jaconi: I added a few comments around this whole block of code. Here's a snipped to show the difference between the approaches (I added and used a third, more readable one):

```shell
#!/usr/bin/env bash

# a)
unset FOO
if [ ! -z ${FOO+x} ]; then
  echo "a) is true"
fi

# b)
FOO=""
if [ ! -z ${FOO+x} ]; then
  echo "b) is true"
fi

# c)
FOO="BAR"
if [ ! -z ${FOO+x} ]; then
  echo "c) is true"
fi

echo "###"

# d)
unset FOO
if [ -n ${FOO+x} ]; then
  echo "d) is true"
fi

# e)
FOO=""
if [ -n ${FOO+x} ]; then
  echo "e) is true"
fi

# f)
FOO="BAR"
if [ -n ${FOO+x} ]; then
  echo "f) is true"
fi

echo "###"

# g)
unset FOO
if [[ -v FOO ]]; then
  echo "g) is true"
fi

# h)
FOO=""
if [[ -v FOO ]]; then
  echo "h) is true"
fi

# i)
FOO="BAR"
if [[ -v FOO ]]; then
  echo "i) is true"
fi
```

dekobon: Wow! That snippet was informative. I learned something new. Thanks for sharing.

dekobon (approved):
LGTM!

PR #72, opened 2022-11-25
Add support for unpriviledged non-root image container

dav-pascual:
Running container images with a non-root user provides additional security against vulnerabilities, and some containers orchestrators such as Openshift enforce this security policy making official regular nginx not compatible out of the box.

These changes, based on ['nginxinc/docker-nginx-unprivileged'](https://github.com/nginxinc/docker-nginx-unprivileged), modify the parent image 'Dockerfile.oss' to allow nginx to execute with a non-root user. 
Basic changes include:
* The default NGINX listen port is now `8080` instead of `80` (this is no longer necessary as of Docker `20.03` but it's still required in other container runtimes).
* The default NGINX user directive in `/etc/nginx/nginx.conf` has been removed.
* The default NGINX PID has been moved from `/var/run/nginx.pid` to `/tmp/nginx.pid`.
* Change `*_temp_path` variables to `/tmp/*`.

### DIFF
--- a/Dockerfile.latest-unpriviledged
+++ b/Dockerfile.latest-unpriviledged
@@ -1,0 +1,21 @@
+# This container images makes the necessary modifications in the
+# inherited image (which could be OSS NGINX or NGINX Plus) in order
+# to allow running NGINX S3 Gateway as a non root user.
+# Steps are based on the official unpriviledged container:
+# https://github.com/nginxinc/docker-nginx-unprivileged/blob/main/Dockerfile-debian.template
+FROM nginx-s3-gateway
+
+# Implement changes required to run NGINX as an unprivileged user
+RUN sed -i "/^server {/a \    listen       8080;" /etc/nginx/templates/default.conf.template \
+    && sed -i '/user  nginx;/d' /etc/nginx/nginx.conf \
+    && sed -i 's,/var/run/nginx.pid,/tmp/nginx.pid,' /etc/nginx/nginx.conf \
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -4,7 +4,7 @@
 
 ## Introduction
 
-This project provides a working configuration of NGINX configured to act as
```
dekobon, line 7: Please revert the whitespace changes in this commit.

Dockerfile.latest-unpriviledged
```diff
@@ -0,0 +1,19 @@
+# This container images makes the necessary modifications in the
+# inherited image (which could be OSS NGINX or NGINX Plus) in order
+# to allow running NGINX S3 Gateway as a non root user
+FROM nginx-s3-gateway
+
+# Implement changes required to run NGINX as an unprivileged user
```
dekobon, line 6: Please add a link in the comments to the official unprivileged container (https://github.com/nginxinc/docker-nginx-unprivileged/blob/main/Dockerfile-debian.template).

dekobon (changes requested):
This looks like a great addition! There are only few small changes needed.

README.md
dav-pascual: Reverted! It was an automatic VSCode thing

Dockerfile.latest-unpriviledged
dav-pascual: Done!

dekobon (approved):

dekobon:
I've incorporated the changes in this PR into this [PR](https://github.com/nginxinc/nginx-s3-gateway/pull/76), so I will close this PR.

PR #76, opened 2022-12-02
Fixes #73

### LINKED ISSUE #73: Container image runs as root and does not work in Openshift
Some Kubernetes distributions like Openshift are more restrictive by default regarding running container with root privileges.
Thus, It would be nice to have an image supporting non-root execution. 
Here is some extended info about the issue https://torstenwalter.de/openshift/nginx/2017/08/04/nginx-on-openshift.html 

I opened a PR adding a Dockerfile that modify parent OSS image with necessary changes to run the container with a non-root user, based on the steps done in https://github.com/nginxinc/docker-nginx-unprivileged/blob/main/Dockerfile-debian.template

https://github.com/nginxinc/nginx-s3-gateway/pull/72

dekobon:
This PR builds upon the previous commits that add support for running unprivileged nginx instances. Here, we normalize spelling, and add CI tooling support for unprivileged nginx instances.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,9 +21,13 @@ jobs:
       - name: Install dependencies
         run: sudo apt-get install -y wait-for-it
       - name: Run tests - latest njs version
-        run: ./test.sh latest-njs-oss
+        run: ./test.sh --latest-njs --type oss
       - name: Run tests - stable njs version
-        run: ./test.sh oss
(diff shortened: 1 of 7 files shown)


dekobon:
Hi @dav-pascual ,

Can you test this PR/branch and see if it works for you with OpenShift? I took your changes and built upon them.

dav-pascual:
@dekobon 
It works like a charm, thanks

PR #85, opened 2023-01-07
Format env variables as table

### LINKED ISSUE #61: Indicate what environment variables are optional in getting started docs
Currently, some optional environment variables are not coded as such. This issue is to improve the documentation by explicitly showing what is required and what isn't.

zacMode:
This documents which env variables are optional (resolves #61 ), based on my recent experience trying to use this. 

I reformatted some of the inline metadata (allowed values, and default values) as separate columns, only using previously provided content. 

Something that was confusing to me when i was first trying to get this running a few days ago, and remains confusing - what does `default` mean when the variable is required? For example, `ALLOW_DIRECTORY_LIST` claims to have a default value of `false`, but i was required to provide it.


dekobon (approved):
Thank you for the contribution!

PR #96, opened 2023-02-25
Bugfix: Handle non-unixtime expiration

MatthewHepburn:
At least in Fargate, the instance role credentials have a human-readable expiration time, not a unix timestamp (for example - `'2023-02-25T19:56:26Z'`).

As a result, we get `NaN` when multiplying this value by 1000 which breaks the expiry checking logic. This leads to credentials being fetched on every request, which is slow and may well exceed the limits of the metadata endpoint.

My proposed solution is to only do the multiplication if our expiration time is numeric, on the assumption that this must be true in some contexts. `Date` can accept the string expiration times directly with no additional processing.

I've only tested this in Fargate - it might be worth ensuring that this still works for credentials in other contexts.

### DIFF
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -988,8 +988,9 @@ async function fetchCredentials(r) {
     }
 
     if (current) {
-        // AWS returns Unix timestamps in seconds, but in Date constructor we should provide timestamp in milliseconds
-        const exp = new Date(current.expiration * 1000).getTime() - maxValidityOffsetMs;
+        // If AWS returns a Unix timestamp it will be in seconds, but in Date constructor we should provide timestamp in milliseconds
+        const expireAt = typeof current.expiration == 'number' ? current.expiration * 1000 : current.expiration
+        const exp = new Date(expireAt).getTime() - maxValidityOffsetMs;
         if (NOW.getTime() < exp) {
             r.return(200);
             return;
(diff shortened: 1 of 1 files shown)


common/etc/nginx/include/s3gateway.js
```diff
@@ -988,8 +988,9 @@ async function fetchCredentials(r) {
     }
 
     if (current) {
-        // AWS returns Unix timestamps in seconds, but in Date constructor we should provide timestamp in milliseconds
-        const exp = new Date(current.expiration * 1000).getTime() - maxValidityOffsetMs;
+        // If AWS returns a Unix timestamp it will be in seconds, but in Date constructor we should provide timestamp in milliseconds
```
dekobon, line 991: Thank you for your contribution. This looks good, but I would like you to write a longer comment that explains the reason for the change in more detail. Also please provide links to the relevant AWS documentation. I hope to merge this soon.

dekobon (changes requested):

common/etc/nginx/include/s3gateway.js
dekobon: Hi Matthew, I was thinking that you add a comment that is basically what you wrote above, but a bit more succinct.

MatthewHepburn: Ah of course! Sorry - I've added an explanatory comment.

I still couldn't find decent documentation for the ECS/Fargate metadata endpoints. But I was able to find better documentation for the EC2 and STS endpoints that are also supported. It appears that neither of these return expiration times as unix timestamps, they both return RFC 3339 strings instead:

- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials
- https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html

So it may be that the multiplication by 1000 is completely redundant. But I'm not proposing to remove it in this MR as there may be some situation I'm not aware of where it does come into play.

PR #99, opened 2023-02-27
fix: Normalize location path to avoid multiple forward slashes & fix docs

### LINKED ISSUE #88: Remove double/multiple forward slashes from the request path
When using the gateway in directory listing mode, if you have a double slash in the path, the request won't find the folder or files.

I suggest removing extra slashes from the path using a rewrite, a map in the config or from JavaScript.

Example:
`https://example.com/pathFoo//file.txt` would become `https://example.com/pathFoo/file.txt`
`https://example.com/pathFoo//subpathFoo///` would become `https://example.com/pathFoo/subpathFoo/`

dav-pascual:
Fixes: https://github.com/nginxinc/nginx-s3-gateway/issues/88 and docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,9 +70,9 @@ Dockerfile.buildkit.plus         Dockerfile with the same configuration as Docke
                                  with support for hiding secrets using Docker's Buildkit
 Dockerfile.latest-njs            Dockerfile that inherits from the last build of the gateway and
                                  then builds and installs the latest version of njs from source
-Dockerfile.latest-unpriviledged  Dockerfiles that inherits from the last build of the gateway and
+Dockerfile.unprivileged  Dockerfiles that inherits from the last build of the gateway and
                                  makes the necessary modifications to allow running the container
-                                 as a non root, unpriviledged user.
+                                 as a non root, unprivileged user.
 settings.example                 Docker env file example
 standalone_ubuntu_oss_install.sh install script that will install the gateway as a Systemd service
 test.sh                          test launcher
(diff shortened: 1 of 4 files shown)


dekobon (changes requested):
Thank you for this PR! The integration tests are failing with the new changes. Can you please modify the integration tests such that they test for the new functionality?

dekobon:
I've been doing some research on how to fix the integration tests and I've been unable to find a method of normalizing slashes using rewrite rules that will preserve the requested schema, hostname, and port. I think we may have to implement this in njs. I did a prototype in njs and it was relatively simple to normalize the uri requested to S3, but it does not result in a HTTP 302 redirect to the client. I'll update as I discover more.

dav-pascual:
I have not taken a a look yet at the integration tests yet, but if the njs solution serves the content correctly even if it doesn't rewrite the URL  / redirect in the client I think that's a valid solution.

In a previous unsuccessful attempt in njs I was getting 404 errors after normalising "uri_path" variable.

dav-pascual:
I tried updating the integration tests to expect a 302 code redirect when there is a double slash, but I am getting a error code return 7 from the run, without much detail

dekobon:
@dav-pascual - I'm traveling right now, so it will be a while until I can engage on this.

violethaze74 (commented):
reviewed

dekobon:
Thank you for this contribution, and apologies for closing it after such a long time. We're closing it as part of a cleanup of stale PRs: the review discussion concluded that rewrite-based slash normalization can't preserve the requested scheme, host, and port, so this needs an njs-based implementation instead, and the docs/Dockerfile.unprivileged rename portion of this PR has already landed separately. The underlying double-slash issue is still real, so a fresh PR taking the njs approach would be very welcome.

Issue #100, opened 2023-03-01
Add temporary security credentials without IAM policy


shawnhankim:
**Is your feature request related to a problem? Please describe.**
As a Software Engineer,
- Currently, I can run `nginx-s3-gateway` using AWS Access key id and secret access key without an IAM Policy.
- In addition to that I want to run and test it either in a container or on a laptop using temporary security credentials which include a session token.

**Describe the solution you'd like**
- Read a session token from the environment variable of `S3_SESSION_TOKEN` if the variable exists and is not empty.
- Otherwise, set to `null` to return credentials which only contain AWS access key ID and secret key.

**Describe alternatives you've considered**
- In case the `S3_SESSION_TOKEN` isn't specified in your AWS profile, the `nginx-s3-gateway` can be accessed using access key id and secret access key as is.

**Additional context**
- This PR is used when a session token is required to access your service without an IAM Policy as the following picture.

<img width="791" alt="credentials-example" src="https://user-images.githubusercontent.com/39923730/222021471-62ba8919-b0b7-41ea-a500-ccc75ec51c3a.png">

dekobon:
Fixed in #101

PR #101, opened 2023-03-01
Add temporary security credentials without IAM policy

shawnhankim:
This PR aims to address #100.

### DIFF
--- a/common/docker-entrypoint.d/00-check-for-required-env.sh
+++ b/common/docker-entrypoint.d/00-check-for-required-env.sh
@@ -52,6 +52,10 @@ else
   required+=("S3_ACCESS_KEY_ID" "S3_SECRET_KEY")
 fi
 
+if [[ -v S3_SESSION_TOKEN ]]; then
+  echo "S3 Session token present"
+fi
+
 for name in ${required[@]}; do
   if [[ ! -v name ]]; then
       >&2 echo "Required ${name} environment variable missing"
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -236,11 +236,13 @@ function _writeCredentialsToFile(credentials) {
  * @returns {undefined|{accessKeyId: (string), secretAccessKey: (string), sessionToken: (string|null), expiration: (string|null)}} AWS instance profile credentials or undefined
  */
 function readCredentials(r) {
-    if (process.env['S3_ACCESS_KEY_ID'] && process.env['S3_SECRET_KEY']) {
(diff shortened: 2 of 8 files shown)


dekobon:
Hi @shawnhankim thank you for the contribution. It seems that some of the integration tests are failing. Would you mind looking into the cause of the failure?

shawnhankim:
Hi @dekobon, 
Thanks for your feedback. Sure thing!

dekobon:
Hi @shawnhankim 
The changes look really good. I've got one request. Can you add a line to the `common/docker-entrypoint.d/00-check-for-required-env.sh` that indicates that `SESSION_TOKEN` is being used?

shawnhankim:
Thanks @dekobon for the great feedback. I have added it. 😃

common/docker-entrypoint.d/00-check-for-required-env.sh
```diff
@@ -49,7 +49,7 @@ elif [[ -v AWS_WEB_IDENTITY_TOKEN_FILE ]]; then
 # If none of the options above is used, require static credentials.
 # See https://docs.aws.amazon.com/sdkref/latest/guide/feature-static-credentials.html.
 else
-  required+=("S3_ACCESS_KEY_ID" "S3_SECRET_KEY")
+  required+=("S3_ACCESS_KEY_ID" "S3_SECRET_KEY" "S3_SESSION_TOKEN")
```
dekobon, line 52: I'm a bit confused about this line. What about the state of the else condition makes the presence of `S3_SESSION_TOKEN`?

Excuse my original comment about this file because I may have been unclear. I was thinking that we would just do something like is at the [bottom of the file](https://github.com/nginxinc/nginx-s3-gateway/blob/master/common/docker-entrypoint.d/00-check-for-required-env.sh#L111).

For example:
```bash
if [[ ! -v "S3_SESSION_TOKEN" ]]; then
  echo "S3 Session token present"
fi
```
(Please check my above conditional syntax for it may be wrong)

Additional, I ran the gateway without `S3_SESSION_TOKEN` (only with `S3_ACCESS_KEY` and `S3_SECRET_KEY`) and it seemed to work fine without triggering a failure state. I'm scratching my head wondering why...

dekobon (changes requested):
Please excuse all of the requests for changes. I think there is one more thing to address.

dekobon (approved):
Looks great.

shawnhankim:
Thanks @dekobon !

common/docker-entrypoint.d/00-check-for-required-env.sh
shawnhankim: Thanks for finding this.

Issue #107, opened 2023-03-03
Support common library to read and write AWS credentials via env variables and cache


shawnhankim:
**Is your feature request related to a problem? Please describe.**
As a Software Engineer,
I want to reuse the common AWS signature library for not only NGINX S3 gateway but also other services such as NGINX Lambda gateway.

As one of the steps,
I want to reuse the library which can read and write AWS credentials via environment variables, a temporary file, or key-value store.

**Describe the solution you'd like**
- Refactor S3 gateway NJS codebase.
- Ability to provide the following reusable functions in `awscredentials.js`:
   - `sessionToken()`: Get the current session token from either the instance profile credential cache or environment variables
   - `readCredentials()`: Read credentials from environment variables, a temporary file, or key-value store.
   - `writeCredentials()`: Write credentials into a temporary file, or key-value store unless environment variables are specified.
- Common utils in `utils.js`:
   - `parseBoolean()`
   - `debug_log()`

**Describe alternatives you've considered**
- The functions of handling credentials can be part of the AWS signature library (e.g., `awssigv4.js`). However, this PR made `awscredentials.js`. Because it is the generic library to support not only signature v4 and future versions.
- NGINX AWS signature repo: will be considered after the common AWS signature library completes.
- NGINX AWS signature module: TBD

**Additional context**
- The following enhancements are separately considered with other PR to avoid any bug based on big changes:
   - Change the name of constants (e.g., `S3_ACCESS_KEY_ID` -> `AWS_ACCESS_KEY_ID`).
   - `fetchCredentials()` w/ EC2, ECS, EKS
   - `signatureV4()`
   - etc.

PR #108, opened 2023-03-03
Common AWS credentials library

shawnhankim:
This PR aims to address #107.

### DIFF
--- a/common/etc/nginx/include/utils.js
+++ b/common/etc/nginx/include/utils.js
@@ -23,6 +23,24 @@ const DEBUG = parseBoolean(process.env['S3_DEBUG']);
 
 
 /**
+ * Parses a string delimited by semicolons into an array of values
+ * @param string {string|null} value representing a array of strings
+ * @returns {Array} a list of values
+ */
+function parseArray(string) {
+    if (string == null || !string || string === ';') {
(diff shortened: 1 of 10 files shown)


common/etc/nginx/include/utils.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+/**
+ * Flag indicating debug mode operation. If true, additional information
+ * about signature generation will be logged.
+ * @type {boolean}
+ */
+const DEBUG = parseBoolean(process.env['S3_DEBUG']);
+
+
```
dekobon, line 24: Why did the `parseArray` function not get moved to the `utils.js` file?

dekobon (approved):
Looks good to me. I only had one comment about a style change.

common/etc/nginx/include/utils.js
shawnhankim: - Thanks @dekobon for your question. 
- Because it is not related to this PR to read and write credentials. So, I was planning to move others to the `utils.js` later to minimize changes per each PR with avoiding any mistakes as much as I can. 
- But, it makes sense to move to the `utils.js`. Hence, I have moved the function in the common utils. 😸

Issue #109, opened 2023-03-06
Support common AWS signature version 2 & 4 library for S3 gateway and other services


shawnhankim:
**Is your feature request related to a problem? Please describe.**
As a Software Engineer,
I want to reuse the common AWS signature version 4 library for not only NGINX S3 gateway but also other services such as NGINX Lambda gateway.

As one of the steps,
I want to reuse the library which creates a signature for use authenticating against AWS compatible APIs.

**Describe the solution you'd like**
- Refactor S3 gateway NJS codebase.
- Ability to provide the following reusable AWS signature V4 functions in `awssig4.js`:
  - `signatureV4()`: Create HTTP Authorization header for authenticating with an AWS compatible v4 API.
  - `_buildCanonicalRequest()`: Creates a canonical request that will later be signed.
  - `_buildSignatureV4()`: Creates a signature for use authenticating against an AWS compatible API.
  - `_buildSigningKeyHash()`: Creates a signing key HMAC. This value is used to sign the request made to the API.
  - `_splitCachedValues()`: Splits the cached values into an array with two elements or returns an empty array if the input string is invalid.
- Ability to provide the following reusable AWS signature V4 functions in `awssig2.js`:
  - `signatureV2()`: Create HTTP Authorization header for authenticating with an AWS compatible v2 API.
- Common utils in `utils.js`:
   - `getAmzDatetime()`: Creates a string in the ISO601 date format (YYYYMMDD'T'HHMMSS'Z').
   - `getEightDigitDate()`: Formats a timestamp into a date string in the format 'YYYYMMDD'.

**Describe alternatives you've considered**
- NGINX AWS signature repo: will be considered once the common AWS signature library completes.
- NGINX AWS signature module: TBD

**Additional context**
- NGINX AWS Signature Library Integration Flow 1 for NGINX S3 Gateway:

<img width="923" alt="nginx-s3-gateway-w-aws-signature" src="https://user-images.githubusercontent.com/39923730/223931950-7c2ff613-e7ec-4f89-9436-0eedcc68b87b.png">

- NGINX AWS Signature Library Integration Flow 2 for NGINX S3 Gateway:

<img width="1563" alt="nginx-s3-gateway-signature-flow" src="https://user-images.githubusercontent.com/39923730/223932053-2eef5ce6-2fc7-48cf-8975-5dc4aa521557.png">

dekobon:
[...]

PR #113, opened 2023-03-09
Common AWS signature version 2 & 4 library for S3 gateway and other services

shawnhankim:
- This PR aims to address #109.
- The doc is added in the 3rd commit.
- The 1st and 2nd commits are same as https://github.com/nginxinc/nginx-s3-gateway/pull/108 and is going to be rebased once it is merged. This is to avoid big changes although it is not the functional breaking change.


dekobon (changes requested):
These changes look good. I would like to see us break out aws v2 sigs into another file like `awssig4.js` so that he project structure is balanced.

shawnhankim:
> These changes look good. I would like to see us break out aws v2 sigs into another file like `awssig4.js` so that he project structure is balanced.

Thanks @dekobon for your review and feedback. I have added AWS Signature V2 Lib in the `awssig2.js`.

dekobon:
Merged in e3068dfbef7619000f7cb06ff5f251022d72d967

PR #116, opened 2023-03-23
Improve documentation around running using EKS IRSA

cablespaghetti:
I had a lot of trouble getting access using IRSA to function correctly using the latest unprivileged container. This PR improves the documentation in the following ways:
* Makes a note about `JS_TRUSTED_CERT_PATH` environment variable required for IRSA to work
* Makes a note about `AWS_REGION` which I had to set in addition to `S3_REGION` for this to work in my environment. Without that I got this error:
```
2023/03/23 17:23:00 [info] 76#76: *10 js: Cached credentials are expired or not present, requesting new ones
2023/03/23 17:23:00 [info] 76#76: *10 js: Could not assume role using web identity: "Missing required AWS_REGION env variable"
2023/03/23 17:23:00 [error] 76#76: *10 auth request unexpected status: 500, client: 10.2.109.58, server: , request: "GET /tails-bootstrap-package/ HTTP/1.1", host: "docs.tailstech.com"
```
* Mentions that it is not possible to use a different uid/gid or set the root filesystem to read-only which we have as the default for apps which we run normally

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -159,6 +159,8 @@ docker run --env-file ./settings --publish 80:8080 --name nginx-s3-gateway \
     nginx-s3-gateway:unprivileged-oss
 ```
 
+It is worth noting that due to the way the startup scripts work, even the unprivileged container will not work with a read-only root filesystem or a specific uid/gid set other then the default of  `101`.
+
 ### Building the NGINX Plus Container Image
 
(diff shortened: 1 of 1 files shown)


dekobon (approved):

dekobon:
Thank you so much for contributing to the documentation!

PR #119, opened 2023-04-14
Fixes #118

### LINKED ISSUE #118: S3_ACCESS_KEY_ID not used when running on an EC2 instance
**Describe the bug**
When running on an EC2 instance with `S3_ACCESS_KEY_ID` and `S3_SECRET_KEY` set, the gateway ignores the credentials set via environment variables and instead attempts to fetch credentials via the instance credentials API.

**To Reproduce**
1. Install Docker on an AWS EC2 instance and configure your S3 bucket.
2. Run the gateway as a docker container and explicitly specify  `S3_ACCESS_KEY_ID` and `S3_SECRET_KEY`.
3. Watch the gateway attempt to get credentials from the instance credential API. *boom*

**Expected behavior**
The gateway should use the specified credentials and not use the instance credentials API. It should only use the instance credentials API if the access key id and secret key are not specified.

dekobon:
The gateway will now use the S3_ACCESS_KEY_ID and S3_SECRET_KEY environment variables when running in an EC2 instance.


shawnhankim (approved):
LGTM! Thanks for finding and fixing it! 👍

ivanitskiy (approved):
Looks good

Issue #111, opened 2023-03-07
Support common library to fetch credentials to generate AWS signature from ECS/EKS/EC2


shawnhankim:
**Is your feature request related to a problem? Please describe.**
As a Software Engineer,
I want to reuse the common AWS signature library for not only NGINX S3 gateway but also other services such as NGINX Lambda gateway.

As one of the steps,
I want to reuse the library which is to fetch credentials to generate AWS signature from the metadata endpoints of EC2, ECS, and EKS.

**Describe the solution you'd like**
- Refactor S3 gateway NJS codebase.
- Ability to provide the following reusable functions in `awscredentials.js`:
   - `fetchCredentials()`: Get the credentials needed to create AWS signatures in order to authenticate to AWS services such as S3 and Lambda.
   - `_fetchEcsRoleCredentials()`: Get the credentials needed to generate AWS signatures from the ECS metadata endpoint.
   - `_fetchWebIdentityCredentials()`: Get the credentials by assuming calling AssumeRoleWithWebIdentity with the environment variable values ROLE_ARN, AWS_WEB_IDENTITY_TOKEN_FILE and HOSTNAME.
   - `_fetchEC2RoleCredentials()`: Get the credentials needed to generate AWS signatures from the EC2 metadata endpoint.

**Describe alternatives you've considered**
- The functions of handling credentials can be part of the AWS signature library (e.g., `awssigv4.js`). However, this PR made `awscredentials.js`. Because it is the generic library to support not only signature v4 and future versions.
- NGINX AWS signature repo: will be considered after the common AWS signature library completes.
- NGINX AWS signature module: TBD

**Additional context**
- N/A

PR #126, opened 2023-05-06
Add payload hash func and default host var; Move fetching credentials func to sig lib

shawnhankim:
To enhance common AWS sig & credentials lib for services in AWS.

**Description:**

- Add a payload hash function to support multiple services in AWS such as S3 and Lambda, and it can be used for not only GET but also POST method.
- Add a NJS var of default host name to be used for Role Session Name at multiple services in AWS
- Move reusable functions of fetching credentials into `awscredentials.js`
- Replace name of env variables of AWS credentials and debug to support common AWS signature lib for multiple services in AWS

**Misc.:**
- Close https://github.com/nginxinc/nginx-s3-gateway/issues/107
- Close https://github.com/nginxinc/nginx-s3-gateway/issues/111


common/etc/nginx/templates/default.conf.template
```diff
@@ -23,6 +24,7 @@ js_var $indexIsEmpty true;
 js_set $s3auth s3gateway.s3auth;
 js_set $awsSessionToken awscredentials.sessionToken;
 js_set $s3uri s3gateway.s3uri;
+js_var $defaultHostName 'nginx-s3-gateway';
```
dekobon, line 27: What's the purpose of setting this variable?

common/etc/nginx/include/s3gateway.js
```diff
@@ -477,234 +442,7 @@ function _requireEnvVars(envVarName) {
...
-/**
- * Get the credentials by assuming calling AssumeRoleWithWebIdentity with the environment variable
- * values ROLE_ARN, AWS_WEB_IDENTITY_TOKEN_FILE and HOSTNAME
- *
- * @returns {Promise<{accessKeyId: (string), secretAccessKey: (string), sessionToken: (string), expiration: (string)}>}
- * @private
- */
-async function _fetchWebIdentityCredentials(r) {
-    const arn = process.env['AWS_ROLE_ARN'];
-    const name = process.env['HOSTNAME'] || 'nginx-s3-gateway';
```
dekobon, line 655: How is the hardcoded value `nginx-s3-gateway` being used? It would be useful to have a comment about it.

dekobon (approved):
Hi @shawnhankim thank you for the solid contribution. My apologies for the delay on the review. This all looks good and I had only a few comments/suggestions.

common/docker-entrypoint.d/00-check-for-required-env.sh
```diff
@@ -34,7 +34,7 @@ required=("S3_BUCKET_NAME" "S3_SERVER" "S3_SERVER_PORT" "S3_SERVER_PROTO"
 if [[ -v AWS_CONTAINER_CREDENTIALS_RELATIVE_URI ]]; then
   echo "Running inside an ECS task, using container credentials"
```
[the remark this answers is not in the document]
shawnhankim: It is a great point. Thanks! Addressed. Let me know if you have any additional comments.

common/etc/nginx/include/awscredentials.js
```diff
@@ -34,24 +68,26 @@ function sessionToken(r) {
...
 
 /**
- * Get the instance profile credentials needed to authenticated against S3 from
+ * Get the instance profile credentials needed to authenticate against S3 from
  * a backend cache. If the credentials cannot be found, then return undefined.
  * @param r {Request} HTTP request object (not used, but required for NGINX configuration)
  * @returns {undefined|{accessKeyId: (string), secretAccessKey: (string), sessionToken: (string|null), expiration: (string|null)}} AWS instance profile credentials or undefined
  */
 function readCredentials(r) {
     // TODO: Change the generic constants naming for multiple AWS services.
```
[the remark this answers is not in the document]
shawnhankim: Thanks for finding this. Addressed!

common/etc/nginx/include/s3gateway.js
shawnhankim: - This is the existing hardcoded value in this repo's codebase. Looks like it is used as a default Role Session Name (`nginx-s3-gateway`) in case `HOSTNAME` is missed.

  - https://github.com/nginxinc/nginx-s3-gateway/blob/master/docs/getting_started.md?plain=1#L54

- As we are moving this function into the common library(`awscredentials.js`), we can consider one of the following options.
  - option 1. remove this default value (`nginx-s3-gateway`) so customers must define the env var (`HOSTNAME`).
  - option 2. replace this value with other variable (e.g. `$defaultHostName`).
  - option 3. any other suggestion?
  - FYI:
    - https://github.com/nginxinc/nginx-s3-gateway/pull/126/files#r1190281902
    - https://github.com/shawnhankim/nginx-s3-gateway/blob/awssig-env/common/etc/nginx/include/awscredentials.js#L373
    - https://github.com/shawnhankim/nginx-s3-gateway/blob/awssig-env/common/etc/nginx/templates/default.conf.template#L27

I was wondering what would be your suggestion.

common/etc/nginx/templates/default.conf.template
shawnhankim: This is the purpose of option 2 in this link: https://github.com/nginxinc/nginx-s3-gateway/pull/126#discussion_r1190622214

common/etc/nginx/include/s3gateway.js
shawnhankim: - I have removed this hardcoded value in this NJS code, because this NJS can be used for not only S3 but also other services such as Lambda.
- I have set this default value in the https://github.com/shawnhankim/nginx-s3-gateway/blob/awssig-env/common/docker-entrypoint.d/00-check-for-required-env.sh#L58 as the following example instead:
  ```bash
   elif [[ -v AWS_WEB_IDENTITY_TOKEN_FILE ]]; then
     echo "Running inside EKS with IAM roles for service accounts"
     if [[ ! -v HOSTNAME ]]; then
       # This environment value is used for Role Session Name. The default value is
       # set as a nginx-s3-gateway unless the value is defined.
       HOSTNAME="nginx-s3-gateway"
     fi
  ```
- Let me know if you have other suggestion, or if you agree with this approach.

common/etc/nginx/templates/default.conf.template
shawnhankim: I have removed this variable, and you could find my answer in the https://github.com/nginxinc/nginx-s3-gateway/pull/126#discussion_r1193013453.

common/etc/nginx/include/s3gateway.js
dekobon: This approach is I believe is the best we can do. I was considering modifying the startup scripts such that if the environment variable `HOSTNAME` is not set that script would run `hostname` and then set `HOSTNAME` to the value returned.

shawnhankim: - Thanks for your opinion and original intent.
- Yeah, I was also confused about the [comment and variable](https://github.com/nginxinc/nginx-s3-gateway/blob/master/docs/getting_started.md?plain=1#L54). So I have replaced `HOSTNAME` with `AWS_ROLE_SESSION_NAME` and added the depreciation message of the variable of `HOSTNAME`.

PR #127, opened 2023-05-10
fix: k8s error exec /docker-entrypoint.sh: exec format error

alejo-jm:
hello, i change the shebang of the shell scripts and now the image run without errors on k8s aws 
error exec /docker-entrypoint.sh: exec format error (mac m1 with FROM --platform=linux/amd64)

Regards

### DIFF
--- a/common/docker-entrypoint.d/00-check-for-required-env.sh
+++ b/common/docker-entrypoint.d/00-check-for-required-env.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 #
 #  Copyright 2020 F5 Networks
 #
--- a/common/docker-entrypoint.sh
+++ b/common/docker-entrypoint.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 #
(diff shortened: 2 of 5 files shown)


dekobon:
Thank you for submitting this issue. Can you tell me more about the environment in which you are having a problem? How could I get access to it? The pattern I'm using to locate bash is a common pattern. With this proposed change, systems that have bash in a directory other than `/bin/bash` will fail to execute.

bijancn:
Mhh. I think I can reproduce the issue. When running a self-built docker image on AWS EKS, I get `exec /docker-entrypoint.sh: exec format error`. My `Dockerfile` looks like this
```
FROM ghcr.io/nginxinc/nginx-s3-gateway/nginx-oss-s3-gateway:latest-20240227 

COPY etc/nginx/templates/default.conf.template /etc/nginx/templates/
```
and I built it on a Mac M1. When I change the docker build command to
```
docker  buildx build --platform linux/amd64 -t registry.jimdo-platform.net/jimdo/openapi-specifications:v5 .
```
I get a different error but at least the entry points seem to get properly loaded.

PR #135, opened 2023-05-26
Added root path prefix in listings template

zc-devs:
Hello.

I list my S3 bucket under some prefix. Currently nginx-s3-gateway doesn't allow it. I can (and I do) strip prefix in HAProxy, but get links without prefix in listing. See diagram below:

nginx-s3-gw-path-prefix

So, this PR prefixes links in directory listing with some predefined value. If it looks good, then I'll add variable into other containers and documentation.


zc-devs:
@dekobon, no problem.

Sure, I can resolve TODOs. Just signal me if we are going with it in general. Does it suit an architecture, etc?

dekobon:
> @dekobon, no problem.
> 
> Sure, I can resolve TODOs. Just signal me if we are going with it in general. Does it suit an architecture, etc?

Yes! This would be a good feature.

zc-devs:
Hi Elijah @dekobon,

Does it look good for you? Could you merge PR?

dekobon:
Excuse my lateness in addressing this PR. I tried to test out the PR, and I saw the [following behavior](https://asciinema.org/a/nWayH975JhSKoI0X3DS4At9Nt). I see the following after enabling `DIRECTORY_LISTING_PATH_PREFIX=main/`

1. The root directory list when accessed at `/` (`http://localhost`) returns an index page that displays as though it was at the `/main` path.
2. If I navigate to a file from the directory listing, I get a `404` because the path to the file is prepended with a `main/` path.

I want to make sure that my understanding of the behavior is correct. I did not understand the behavior well based on the text in the readme.

To confirm, this change modifies the paths presented by the index page, but does not change the actual paths served. Is that correct?

zc-devs:
Yes, you absolutely right.

I make requests through another proxy (in my case via HAProxy), which removes the same prefix from request path:
`http-request set-path %[path,regsub(^/main,/)]`

dekobon:
I've made some changes to your PR and pushed it as a branch here: https://github.com/dekobon/nginx-s3-gateway/tree/dir-list-prefix

In particular, I removed the templating done on the includes directory and instead used an xlst parameter.

Please take a look at the above branch, test with your configuration and let me know the results.

If it works for you, then I'd like to merge it.

zc-devs:
Hi.
Your fixes work. Let's merge it.
Thank you.

dekobon:
Changes merged in commits:
e4c95c41c9e03c65adcefb5f1303bad275169cca
54653a86932dee857482972c77dcf9e4705f58d7
8b57860ff8633151565c032433b7e29ecf6c68b0
f491aaf6ecc1ffb8075f3d43895862344af47072

PR #136, opened 2023-05-30
Periodic DNS resolution

onlynone:
This should allow both the oss and plus versions to do periodic DNS resolution by using a variable in the `proxy_pass` and eliminating the jump through an upstream.


dekobon:
Thank you for taking the time to put together this PR.

Unfortunately, there are some invisible side-effects of this change:
* Loss of keep-alive connections
* Loss of statistics for upstream

Also, DNS resolution for OSS is currently in [review](https://mailman.nginx.org/pipermail/nginx-devel/2023-February/4MCLSVRK7EX6DNKHFZN6CA4SKZUSA3GA.html), so first-class support for it should be coming in the near term.

Out of the above issue, the loss of keep-alive connections is a significant drawback for the S3 use case. Please let me know what your thoughts are.

onlynone:
> Also, DNS resolution for OSS is currently in [review](https://mailman.nginx.org/pipermail/nginx-devel/2023-February/4MCLSVRK7EX6DNKHFZN6CA4SKZUSA3GA.html), so first-class support for it should be coming in the near term.

That would be great. I hope that feature gets merged soon.

> Out of the above issue, the loss of keep-alive connections is a significant drawback for the S3 use case. Please let me know what your thoughts are.

Can you explain how significant the keep-alive connections are with S3? Is there no way to get keep-alive connections without going through an upstream?

dekobon:
DNS resolution has been open sourced: https://blog.nginx.org/blog/dynamic-dns-resolution-open-sourced-in-nginx

Closing in favor of changes that use the OSS approach

PR #159, opened 2023-08-04
Add missing CORS headers for 404 status code

Fr1z2r:
When file not exist in s3 nginx return 404 response without CORS headers.


dekobon (approved):
Thank you for the contribution!

PR #162, opened 2023-08-08
Fix error messages for missing dependencies, allow test suite to run on MacOS by supporting `md5`

4141done:
# What
Fix error messages for missing dependencies, allow test suite to run on MacOS by supporting `md5`

## Dependency Checks
There were some checks like this:

```bash
curl_cmd="$(command -v curl)"
if ! [ -x "${curl_cmd}" ]; then
  e "required dependency not found: curl not found in the path or not executable"
  exit ${no_dep_exit_code}
fi
```

However, `command -v` exits with an error if the command does not exist. This results in the nice error message not being shown.

The solution is to avoid the assignment of the variable erroring and instead relying on the executability check to make sure we have the right thing

## `md5` compatibility
The test script depends on `md5sum` which is not available on MacOS. Thus, tests would fail at the start when run on MacOS.
MacOS has a tool called `md5`, which when called with the `-r` flag which reverses the format of the output.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -55,7 +55,7 @@ additional modules.
 
 ## Testing
 
-Automated tests require `docker`, `docker-compose`, `curl`, and `mc` (the minio [cli tool](https://github.com/penpyt/asdf-mc)) to be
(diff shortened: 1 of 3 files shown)


dekobon (approved):
This looks good to me.

PR #167, opened 2023-08-17
Fixes #164

### LINKED ISSUE #164: With Directory listing turned on query parameters cause 500 errors
**Describe the bug**
With Directory listing turned on query parameters cause 500 errors

**To Reproduce**
1. Start container with APPEND_SLASH_FOR_POSSIBLE_DIRECTORY=true and ALLOW_DIRECTORY_LIST=true with an S3 bucket with an empty index.html at `/test/index.html`
2. Run the container and navigate to /test/?asdf=asdf
3. 500 error at s3PreListing

**Expected behavior**
Should show the content of index.html

**Your environment**
Container is the latest release
Deployed on AWS EC2 with EKS with AWS S3 as the backend

dekobon:
This PR fixes the issue #164 

## Expectations
This solution assumes that the following configuration options are set to true
* `APPEND_SLASH_FOR_POSSIBLE_DIRECTORY`
* `PROVIDE_INDEX_PAGE`
* `ALLOW_DIRECTORY_LIST`

Given a folder `test` **with** an `index.html` present in the directory, the `index.html` should be served for:
* `/test`
* `/test/`
* `/test?foo=bar`
* `/test/?foo=bar`

Given a folder `test` **WITHOUT** an `index.html` present in the directory, files in the directory should be listed for:
* `/test`
* `/test/`
* `/test?foo=bar`
* `/test/?foo=bar`

## Notes
* The `@trailslash` was rewriting to `$request_uri` with a trailing slash on the end.  In the case of `/test?foo=bar` we'd wind up with `/test?foo=bar/` which when combined with the other changes led to a rewrite loop
* Changed the check for directory or file to consider the path without anchor or querystring
* I added yet another integration test suite and shuffled around the conditionals that maybe make the tests even more confusing - but do cover this case.


4141done:
I think your changes here look good, but doing some testing I found some odd behavior.  I'm not 100% sure of all my conclusions, but I put together a reference draft PR for your reference. https://github.com/nginxinc/nginx-s3-gateway/pull/169

4141done (approved):
@dekobon I think these changes will accomplish the fix for issue https://github.com/nginxinc/nginx-s3-gateway/issues/164

We have not received confirmation from the original reporter but some time has passed and this certainly a bug so I think we can move ahead with the fix.

If you could just look over my removal of the debug code and give a 👍 then I think we can merge.

Issue #165, opened 2023-08-17
Add JSDoc type definitions for internal NGINX types


dekobon:
If the internal NGINX types were defined using JSDoc `@typedef`, developers writing njs would get a nice auto-completing workflow when writing njs. This issue is a feature request to add the requisite JSDoc.

dekobon:
Done in #177 (2134898): JSDoc annotations referencing the NGINX request/types were added throughout the njs sources, with the actual type definitions supplied by the official `njs-types` package (a devDependency) rather than hand-written `@typedef` blocks, and JSDoc site generation is wired up via `make jsdoc`. Closing as implemented.

PR #168, opened 2023-08-17
Issue/165

dekobon:
This change adds support for building JSDoc docs using nodejs. It also improves the JSDoc annotations used in the njs source code.

The rationale for this change is that by improving JSDocs, you improve the static code analysis experience when using an IDE to work on njs code.

This change incorporates the following:
* Adds `package.json` which references the TypeScript types for internal njs types, jsdoc, and jsdoc dependencies (taffydb)
* Reworks and improves the existing JSDoc annotations so that they reference the TypeScript types
* Adds a Makefile directive to execute jsdoc


README.md
```diff
@@ -68,6 +68,7 @@ deployments/                     contains files used for deployment technologies
 docs/                            contains documentation about the project
 examples/                        contains additional `Dockerfile` examples that extend the base 
                                  configuration
```
4141done, line 70: Are there any notes we should make about editor integration or docs generation? Do we need to add a note about the nodejs requirement?

GNUmakefile
```diff
@@ -22,3 +22,13 @@ help:
 .PHONY: test
 test: ## Run all tests
 	$Q $(CURDIR)/test.sh --type oss --unprivileged false --latest-njs false
+
+out:
```
4141done, line 26: Two comments:
1. You can probably use `npx jsdoc -c` rather than directly referencing the js file which could change out from under us.
2. `out` seems like a rather generic name for building the docs.  `docs` might be more intuitive?

common/etc/nginx/include/awscredentials.js
```diff
@@ -33,10 +47,16 @@ const NOW = new Date();
 const ECS_CREDENTIAL_BASE_URI = 'http://169.254.170.2';
 
 /**
+ * URL to EC2 Instance Metadata Service (IMDS) token endpoint
+ * @see {@link https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html| EC2 Instance Metadata Service}
```
4141done, line 51: Love the usage of `@see`

common/etc/nginx/include/s3gateway.js
```diff
@@ -322,9 +358,9 @@ function redirectToS3(r) {
 
     if (isDirectoryListing && (r.method === 'GET' || r.method === 'HEAD')) {
         r.internalRedirect("@s3PreListing");
-    } else if ( PROVIDE_INDEX_PAGE == true ) {
+    } else if (PROVIDE_INDEX_PAGE === true) {
```
4141done, line 361: These changes seem like they are from [this PR](https://github.com/nginxinc/nginx-s3-gateway/pull/167). Might want to remove them from or note that this PR is dependent on that one being merged first and a rebase on this one.

4141done (approved):
🐳 👍 🐳 
This looks good to me. I have minor comments but nothing blocking other than the stray code from https://github.com/nginxinc/nginx-s3-gateway/pull/167

I think we can provide some more affordances around the nodejs dependency such as the option of docker based runs for node-based tooling but since this is not a requirement I think it can be done in a later PR.

One caveat is that I don't use Typescript often and I don't use an editor that integrates with these annotations so I can't speak for the effectiveness/correctness of the actual editor stuff.

ivanitskiy:
added some changes and tested in VSC. and submitted a PR https://github.com/dekobon/nginx-s3-gateway/pull/1 to share my findings

4141done:
> added some changes and tested in VSC. and submitted a PR [dekobon#1](https://github.com/dekobon/nginx-s3-gateway/pull/1) to share my findings

I would suggest that we add linting and formatting each as separate changes. I have opened a discussion on linting and formatting [here](https://github.com/nginxinc/nginx-s3-gateway/discussions/163) if you have opinions.

dekobon:
Closing as superseded: the core of this change (JSDoc annotations across the njs sources, jsdoc generation config, and npm packaging with njs-types) landed in #177. The remaining lint/format tooling scope can be tracked against the still-open #165.

PR #169, opened 2023-08-18
[Je issue/164

4141done:
# What
This PR presents some additional ideas to be considered for addition to https://github.com/nginxinc/nginx-s3-gateway/pull/167 aiming to fix issue https://github.com/nginxinc/nginx-s3-gateway/issues/164

## Expectations
This solution assumes that the following configuration options are set to true
* `APPEND_SLASH_FOR_POSSIBLE_DIRECTORY`
* `PROVIDE_INDEX_PAGE`
* `ALLOW_DIRECTORY_LIST`

Given a folder `test` **with** an `index.html` present in the directory, the `index.html` should be served for:
* `/test`
* `/test/`
* `/test?foo=bar`
* `/test/?foo=bar`

Given a folder `test` **WITHOUT** an `index.html` present in the directory, files in the directory should be listed for:
* `/test`
* `/test/`
* `/test?foo=bar`
* `/test/?foo=bar`

## Notes
* The `@trailslash` was rewriting to `$request_uri` with a trailing slash on the end.  In the case of `/test?foo=bar` we'd wind up with `/test?foo=bar/` which when combined with the other changes led to a rewrite loop
* Changed the check for directory or file to consider the path without anchor or querystring
* I added yet another integration test suite and shuffled around the conditionals that maybe make the tests even more confusing - but do cover this case.

### DIFF
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -333,8 +333,12 @@ function redirectToS3(r) {
 
 function trailslashControl(r) {
     if (APPEND_SLASH) {
+        // For the purposes of understanding whether this is a directory,
+        // consider the uri without query params or anchors
+        const path = r.variables.uri_path.split(/[?#]/)[0];
+
         const hasExtension = /\/[^.\/]+\.[^.]+$/;
-        if (!hasExtension.test(r.variables.uri_path)  && !_isDirectory(r.variables.uri_path)){
+        if (!hasExtension.test(path)  && !_isDirectory(path)){
             return r.internalRedirect("@trailslash");
         }
     }
@@ -447,6 +451,8 @@ function _escapeURIPath(uri) {
  * @private
  */
 function _isDirectory(path) {
+    // if (!path) return false;
+    // str.slice(-1);
     if (path === undefined) {
         return false;
     }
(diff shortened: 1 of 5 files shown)


common/etc/nginx/include/s3gateway.js
```diff
@@ -449,16 +451,9 @@ function _escapeURIPath(uri) {
...
 function _isDirectory(path) {
-    if (path === undefined) {
-        return false;
-    }
-    const len = path.length;
-
-    if (len < 1) {
-        return false;
-    }
+    if (!path) return false;
```
4141done, line 454: **Annotation**:
Not a big deal, but the previous implementation didn't account for `null`.  Perhaps that was on purpose? Either way, if we assume `null`, `undefined`, or `""` are all "not a directory" this function gets a lot simpler.

common/etc/nginx/templates/default.conf.template
```diff
@@ -326,7 +326,7 @@ server {
 
     location @trailslash {
         # 302 to request without slashes
-        rewrite ^ $scheme://$http_host$request_uri/ redirect;
+        rewrite ^ $scheme://$http_host$uri/$is_args$query_string redirect;
```
4141done, line 329: **Annotation**:
Insert the trailing slash after the path and preserve the query string after it.

4141done:
Closing in favor of https://github.com/nginxinc/nginx-s3-gateway/pull/167

PR #172, opened 2023-09-13
Fixes #171

### LINKED ISSUE #171: nginx: [emerg] invalid IPv6 address in resolver "[NOTE:]" in /etc/nginx/conf.d/upstreams.conf:2
**Describe the bug**
Poorly constrained `grep` in the bash scripts seems to pick up a comment from the `/etc/resolv.conf` file.

**To Reproduce**
1. Start container ```docker run --env-file ./settings --publish 8011:80 --name nginx-s3-gateway nginx-s3-gateway:oss```
2. View logs
3. See error:
```
2023/09/13 11:50:22 [emerg] 1#1: invalid IPv6 address in resolver "[NOTE:]" in /etc/nginx/conf.d/upstreams.conf:2
nginx: [emerg] invalid IPv6 address in resolver "[NOTE:]" in /etc/nginx/conf.d/upstreams.conf:2
```

**Expected behavior**
The server starts as expected.

**Your environment**
* Local clone of the master branch at time of writing.

**Additional context**
See this line beginning with `# NOTE` and ending in `nameservers`...
```
nameserver 8.8.4.4
nameserver 94.198.184.14
nameserver 94.198.184.34
# NOTE: the libc resolver may not support more than 3 nameservers.
# The nameservers listed below may not be recognized.
nameserver 8.8.8.8
```
 ... it qualifies for the `grep` commands found here: https://github.com/search?q=repo%3Anginxinc%2Fnginx-s3-gateway%20%22grep%20nameserver%22&type=code
**Sensitive Information**

dekobon:
This fix should strip comments from `resolv.conf` files before they are parsed by grep. This change assumes that grep and sed are the GNU variants.


common/docker-entrypoint.sh
```diff
@@ -35,7 +35,7 @@ if [ -z ${DNS_RESOLVERS+x} ]; then
   resolvers=""
   for ip in $(grep nameserver /etc/resolv.conf | cut -d' ' -f2 | xargs)
   do
-    if echo "${ip}" | grep -q ':'; then
+    if echo "${ip}" | sed '/^[[:blank:]]*#/d;s/#.*//' | grep -q ':'; then
```
4141done, line 38: Per the comment on the issue, did you look at https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/15-local-resolvers.envsh for handling this?
It seems that in the case where `DNS_RESOLVERS` is not specified, we could potentially set it equal to the value of the env var exported by this script which I think should already have been run by this point?

4141done (changes requested):
This change looks like it contains a lot more than the stated fix. Can we trim it down to the relevant changes? Also left a small comment on the relevant portion to check if we've explored a path suggested by the reporter.

common/docker-entrypoint.sh
dekobon: Oh, dang, I branched from the wrong commit. I'm going to redo this commit.

4141done:
Closing in favor of https://github.com/nginxinc/nginx-s3-gateway/pull/176 which does the same thing

Issue #175, opened 2023-10-02
Improvement: Properly incorporate docker entrypoint scripts from the official docker image


4141done:
Current behavior:
It appears the [the docker entrypoint scripts ](https://github.com/nginxinc/docker-nginx/tree/master/entrypoint) files may not be getting hooked as we would expect.

Desired behavior:
The standard entrypoint files should be hooked except where they should be overridden by this project's scripts or where the environment variables in this project would clash or become confusing.

dekobon:
Done in #449 (31efb1a): the project's custom `common/docker-entrypoint.sh` — which wholesale replaced the official image's entrypoint — was removed. All image variants now rely on the official nginx image's entrypoint machinery unchanged (including its template processing) and only add drop-in scripts under `/docker-entrypoint.d/`; the formerly forked logic (resolver detection, defaults) lives in those hook scripts. Closing as implemented.

Issue #83, opened 2022-12-29
Generated nginx config is invalid if namesevers are IPv6-only


pcorpet:
**Describe the bug**
This is the same bug as https://github.com/nginx-proxy/nginx-proxy/issues/938

Currently, if there are IPv6 nameservers in `/etc/resolv.conf`, they are not properly formatted when they are inserted into the Nginx template. Nginx requires them to be surrounded by `[]` like `[2a02::1:1]`.

**To Reproduce**
I'm not sure how to manage a `resolv.conf` inside the container to have an IPv6 nameserver (it's the case in my setup using an EKS cluster, probably with some IPv6 settings).

1. Start container with a line `nameserver: 2a02::1:1` in `/etc/resolv.conf`
2. Check logs

```
[emerg] 1#1: invalid port in resolver "2a02::1:1" in /etc/nginx/conf.d/upstreams.conf:2
```

**Expected behavior**

No error on load up.

**Your environment**
* Version of the repo: ghcr.io/nginxinc/nginx-s3-gateway/nginx-oss-s3-gateway:latest-20221216
* S3 backend implementation you are using (AWS, Ceph, NetApp StorageGrid, etc): AWS
* How you are deploying: Helm (kubernetes)
* NGINX type (OSS/Plus): OSS
* Authentication method (IAM, IAM with Fargate, IAM with K8S, AWS Credentials, etc): IAM with K8S

PR #176, opened 2023-10-02
fix: gracefully handle comments in resolv.conf. Fixes #171

### LINKED ISSUE #171: nginx: [emerg] invalid IPv6 address in resolver "[NOTE:]" in /etc/nginx/conf.d/upstreams.conf:2

4141done:
# What
Fixes an issue where comments in the `/etc/resolv.conf` would get into the dns resolvers list and cause an error when starting the s3 gateway.

## How
Following a suggestion on the issue, the logic from [this file](https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/15-local-resolvers.envsh) in the official NGINX docker container was ported over to replace the existing logic.

Given a `/etc/resolv.conf` that looks like this:
```
nameserver 127.0.0.11
nameserver 8.8.4.4
nameserver 94.198.184.14
nameserver 94.198.184.34
# NOTE: the libc resolver may not support more than 3 nameservers.
# The nameservers listed below may not be recognized.
nameserver 8.8.8.8
options ndots:0

```

The existing code would produce this:
```
127.0.0.11 8.8.4.4 94.198.184.14 94.198.184.34 [NOTE:] The 8.8.8.8
```

With this change applied it looks like this:
```
127.0.0.11 8.8.4.4 94.198.184.14 94.198.184.34 8.8.8.8
```

The startup printout looks like this with the change applied:
```
Origin: http://bucket-1.minio:9000
Region: us-east-1
Addressing Style: virtual
AWS Signatures Version: v2
DNS Resolvers: 127.0.0.11 8.8.4.4 94.198.184.14 94.198.184.34 8.8.8.8
Directory Listing Enabled: false
Directory Listing Path Prefix:
Provide Index Pages Enabled:
Append slash for directory enabled:
Stripping the following headers from responses: x-amz-;
CORS Enabled: 0
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
```

## Notes
Ideally we should be incorporating the entrypoint scripts from the official nginx docker image.  For now we're just porting key logic to get the issue solved quickly.  The integration of these scripts will have some other concerns.  An issue has been filed for this work: https://github.com/nginxinc/nginx-s3-gateway/issues/175

### DIFF
--- a/common/docker-entrypoint.sh
+++ b/common/docker-entrypoint.sh
@@ -32,16 +32,11 @@ parseBoolean() {
 
 # This line is an addition to the NGINX Docker image's entrypoint script.
 if [ -z ${DNS_RESOLVERS+x} ]; then
-  resolvers=""
-  for ip in $(grep nameserver /etc/resolv.conf | cut -d' ' -f2 | xargs)
-  do
-    if echo "${ip}" | grep -q ':'; then
-      resolvers="$resolvers [${ip}]"
-    else
-      resolvers="$resolvers $ip"
-    fi
-  done
-  export DNS_RESOLVERS="${resolvers}"
+   # This method of pulling individual nameservers from
+   # /etc/resolv.conf taken from the entrypoint script in the
+   # official docker image.
+   # https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/15-local-resolvers.envsh
+  export DNS_RESOLVERS="$(awk 'BEGIN{ORS=" "} $1=="nameserver" {print $2}' /etc/resolv.conf)"
 fi
 
 # Normalize the CORS_ENABLED environment variable to a numeric value
--- a/common/docker-entrypoint.sh
+++ b/common/docker-entrypoint.sh
@@ -32,11 +32,21 @@ parseBoolean() {
 
 # This line is an addition to the NGINX Docker image's entrypoint script.
 if [ -z ${DNS_RESOLVERS+x} ]; then
-   # This method of pulling individual nameservers from
-   # /etc/resolv.conf taken from the entrypoint script in the
-   # official docker image.
-   # https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/15-local-resolvers.envsh
-  export DNS_RESOLVERS="$(awk 'BEGIN{ORS=" "} $1=="nameserver" {print $2}' /etc/resolv.conf)"
+    resolvers=""
+
+    # This method of pulling individual nameservers from
+    # /etc/resolv.conf taken from the entrypoint script in the
+    # official docker image.
+    # https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/15-local-resolvers.envsh
+    for ip in $(awk 'BEGIN{ORS=" "} $1=="nameserver" {print $2}' /etc/resolv.conf)
+    do
+        if echo "${ip}" | grep -q ':'; then
+          resolvers="$resolvers [${ip}]"
+        else
+          resolvers="$resolvers $ip"
+        fi
+    done
+    export DNS_RESOLVERS="${resolvers}"
 fi
 
 # Normalize the CORS_ENABLED environment variable to a numeric value


ciroque (approved):
I like what you've done here

common/docker-entrypoint.sh
```diff
@@ -32,16 +32,11 @@ parseBoolean() {
...
-    else
-      resolvers="$resolvers $ip"
-    fi
-  done
-  export DNS_RESOLVERS="${resolvers}"
+   # This method of pulling individual nameservers from
+   # /etc/resolv.conf taken from the entrypoint script in the
+   # official docker image.
+   # https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/15-local-resolvers.envsh
+  export DNS_RESOLVERS="$(awk 'BEGIN{ORS=" "} $1=="nameserver" {print $2}' /etc/resolv.conf)"
```
dekobon, line 39: Does this awk pattern support ipv6 syntax? In order to template in ipv6 ips into nginx config, they must be enclosed in `[]`. This was raised in [this bug](https://github.com/nginxinc/nginx-s3-gateway/issues/83).
The fix was [here](https://github.com/nginxinc/nginx-s3-gateway/commit/97a87377bdda5de8d61442d81514cd61ed48ee32).

dekobon (changes requested):
No changes requested, rather this is a request to make sure that the new method does not cause a regression with ipv6 addresses.

common/docker-entrypoint.sh
4141done: Thank you for catching this! I had overlooked this case and missed the enclosure in square brackets from the original code.  I've made changes in [this commit](https://github.com/nginxinc/nginx-s3-gateway/pull/176/commits/0c1d73234c387f2a47f4e2150da3ae7595b3213e) that combines your fix with the changes from the official dockerfile entrypoint.

Given a `resolv.conf` that looks like this:
```
nameserver 127.0.0.11
nameserver 8.8.4.4
nameserver 94.198.184.14
nameserver 94.198.184.34
# NOTE: the libc resolver may not support more than 3 nameservers.
# The nameservers listed below may not be recognized.
nameserver 8.8.8.8
# This is another comment
nameserver 8.8.8.9
nameserver fe80::f03c:93ff:fe25:6762
nameserver 8.9.9.9
nameserver 2a03:a800:2:1::1
nameserver 8.9.9.7
options ndots:0
```

We see:
```
DNS Resolvers:  127.0.0.11 8.8.4.4 94.198.184.14 94.198.184.34 8.8.8.8 8.8.8.9 [fe80::f03c:93ff:fe25:6762] 8.9.9.9 [2a03:a800:2:1::1] 8.9.9.7
```

Which looks right to me. Based on my read of the `resolver` syntax each address should be separately enclosed in square brackets.

dekobon (approved):

PR #181, opened 2023-10-16
feature: add CORS_ALLOW_PRIVATE_NETWORK_ACCESS env var

### LINKED ISSUE #180: Handle `Access-Control-Request-Private-Network: true` CORS headers
**Is your feature request related to a problem? Please describe.**

We are using nginx-s3-gateway to serve sourcemaps for production services. The s3 gateway lives on a private AWS VPC and is only available over our VPN. This means that users on the VPN can get sourcemaps attached to to compiled code running in production, while customers can not.

Recently Chrome has made an update to prevent unintentional Private Network Access.

When Chrome makes a request for sourcemaps to servers using private IP address space (10.0.0.0/8, etc.), it now sends a CORS OPTIONS preflight with the header `Access-Control-Request-Private-Network: true`, and expects to get a response `Access-Control-Allow-Private-Network: true`.

**Describe the solution you'd like**

There are two possible solutions I can see, though there might be more:

* Add an environment variable to configure arbitrary headers, e.g. `RESPONSE_HEADERS_TO_ADD` or `HEADERS_TO_ADD`
* Add an environment variable just for private network access, e.g. `CORS_ALLOW_PRIVATE_NETWORK_ACCESS` or `CORS_ALLOW_PRIVATE_NETWORK`.

**Describe alternatives you've considered**

I could fork this repo and maintain it myself, but this seems like a valuable usecase.

**Additional context**

https://developer.chrome.com/blog/private-network-access-preflight/
https://developer.chrome.com/blog/private-network-access-update/

danielcompton:
Fixes #180.


common/etc/nginx/templates/gateway/cors.conf.template
```diff
@@ -1,4 +1,5 @@
 set $request_cors "${request_method}_${CORS_ENABLED}";
+set $cors_private_network "${CORS_ALLOW_PRIVATE_NETWORK_ACCESS};
```
4141done, line 2: Missing `"` before the `;` is what is causing the current test failures.

common/etc/nginx/templates/gateway/cors.conf.template
```diff
@@ -11,6 +12,18 @@ if ($request_cors = "OPTIONS_1") {
...
+    # Allow/deny Private Network Access CORS requests.
+    # https://developer.chrome.com/blog/private-network-access-preflight/
+    #
+    if ($http_access_control_request_private_network = "true") {
+        if ($cors_private_network = "true") {
+            add_header 'Access-Control-Allow-Private-Network' 'true';
+        }
+        if ($cors_private_network = "false") {
+            add_header 'Access-Control-Allow-Private-Network' 'false';
+        }
```
4141done, line 25: Unfortunately this code is not going to work since the `if` directive can't be used in the `if` context (no nested `if`s).

If we decide that it's OK not to make the `CORS_ALLOW_PRIVATE_NETWORK_ACCESS` configuration object dependent on `CORS_ENABLED` and trust users to enable both themselves, then we can write something like:

```
set $cors_private_network "${CORS_ALLOW_PRIVATE_NETWORK_ACCESS}";

if ($cors_private_network = 'true') {
    set $allow_cors_private_network $http_access_control_request_private_network;
}

if ($cors_private_network = 'false') {
    set $allow_cors_private_network 'false';
}

if ($request_cors = "OPTIONS_1") {
    # ...
    # Allow/deny Private Network Access CORS requests.
    # https://developer.chrome.com/blog/private-network-access-preflight/
    #
    add_header 'Access-Control-Allow-Private-Network' $allow_cors_private_network;

```

This would result in the following behavior:

- If `CORS_ALLOW_PRIVATE_NETWORK_ACCESS` is set to `"true"`, the `Access-Control-Allow-Private-Network` would be sent in the response with whatever the value of the `Access-Control-Request-Private-Network` was.  That maybe creates the potential for strange bugs so maybe just having it be `"true"` might be better?
- If `CORS_ALLOW_PRIVATE_NETWORK_ACCESS` is set to `"true"`, the `Access-Control-Allow-Private-Network` would be sent in the response with the value `"false"`
- If `CORS_ALLOW_PRIVATE_NETWORK_ACCESS` is not specified, no header will be sent `add_header` won't send the header if given a falsey value).

It would be possible switch on the presence of the header and its value by using a `map` directive in the `default.conf` file but I'm a little hesitant to go there because it introduces a lot of complexity for a specific use case that is still in the proposal stage.

Interested to hear your thoughts.

docs/getting_started.md
```diff
@@ -40,6 +40,7 @@ running as a Container or as a Systemd service.
...
 | `CORS_ALLOWED_ORIGIN`                 | No        |                              |           | value to set to be returned from the CORS `Access-Control-Allow-Origin` header. This value is only used if CORS is enabled. (default: \*)                                                                                                                                                                                                                                                                                                                    |
+| `CORS_ALLOW_PRIVATE_NETWORK_ACCESS`   | No        | `true`, `false`              |           | Flag that enables responding to the CORS OPTIONS pre-flight request header `Access-Control-Request-Private-Network` with the `Access-Control-Allow-Private-Network` header. If the value is "true", responds with "true", if "false" responds with "false. If the environment variable is blank/not set, does not respond with any header. This value is only used if CORS is enabled. See [Private Network Access: introducing preflights](https://developer.chrome.com/blog/private-network-access-preflight/) for more information about this header. |
```
4141done, line 43: Thank you for documenting this so well.

4141done (commented):
So after looking through this more closely in the review it seems that implementing this using the logic  in this PR will be a little more complicated than I thought.

To enumerate some of the options:

1. You could maintain your own modification to the `common/etc/nginx/templates/gateway/cors.conf.template` file including the header.  We could document how to do this in the README.
2. We could explore adding the ability to always include certain response headers. That would result in `Access-Control-Allow-Private-Network: true` being sent always but might be a good general purpose improvement.
3. We could implement your proposed `CORS_ALLOW_PRIVATE_NETWORK_ACCESS` as I suggested in my comment (always returning `true` if enabled). But without any awareness of the actual headers received.

These are all options that I am fine going with but you know the needs of someone with this use case better than I do.  I don't feel comfortable putting in a lot of logic to set the headers conditionally but I could certainly supply that code if you wanted to implement that in your deployment.

common/etc/nginx/templates/gateway/cors.conf.template
danielcompton: > It would be possible switch on the presence of the header and its value by using a map directive in the default.conf file but I'm a little hesitant to go there because it introduces a lot of complexity for a specific use case that is still in the proposal stage.

Yeah, I agree that it's not worth burning too much complexity for this.

> We could implement your proposed CORS_ALLOW_PRIVATE_NETWORK_ACCESS as I suggested in my comment (always returning true if enabled). But without any awareness of the actual headers received.

I think this is a good option. From my reading of the spec, there should be no issues with returning this header all the time. The [spec](https://wicg.github.io/private-network-access/) only contemplates a single value of true being checked for as the successful response value.

Do you want me to update the PR? I'm happy for you to if you'd like as well.

4141done: Yeah if you don't mind can you please update the PR and verify that the code still works for your use case that would be awesome.

The test suite is a little harder to update so I'd be happy to add tests for you. If you [allow me to push](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) I can do that once your updates are in.

danielcompton:
@4141done sorry for the delay on this. I've updated the code, tested that it works locally, and also updated some other parts of the Dockerfiles and scripts that I think also needed to be updated.

alessfg:
Hi there @danielcompton! We have changed the base branch of the repo since your original commits -- I edited the PR to target the new base branch (`main`) but it looks like there are some merge conflicts. Can you update the PR when you get a chance?

danielcompton:
Hi @alessfg sorry for the delay, I've rebased this now and resolved the merge conflicts.

danielcompton:
recheck

alessfg:
Thanks! I think you might need to agree to the CLA on a separate message (sorry about this, we are still testing out the workflow!)

4141done:
Thank you for picking this up again @danielcompton 🎉 

Let me rebuild my context on this issue and review this week.

danielcompton:

4141done:
Ok so I spoke to some of my more knowledgeable NGINX colleagues and there was some concern about introducing a `set` in an `if` since there are[ known issues with if in a location](https://www.f5.com/company/blog/nginx/avoiding-top-10-nginx-configuration-mistakes#if). We need to fix the existing usages in this file but probably should not introduce a new one.

Luckily we already have an area in the bootstrap script for this project where we message the cors-related variables so my suggestion is that we put this there for now.  Here's a diff that illustrates this:

```bash
diff --git a/common/docker-entrypoint.sh b/common/docker-entrypoint.sh
index 8688602..939c57b 100644
--- a/common/docker-entrypoint.sh
+++ b/common/docker-entrypoint.sh
@@ -68,6 +68,12 @@ if [ -z "${CORS_ALLOWED_ORIGIN+x}" ]; then
   export CORS_ALLOWED_ORIGIN="*"
 fi
 
+# See documentation for this feature. We do not parse this as a boolean
+# since "true" and "false" are the required values of the header this populates
+if [ "${CORS_ALLOW_PRIVATE_NETWORK_ACCESS}" != "true" ] && [ "${CORS_ALLOW_PRIVATE_NETWORK_ACCESS}" != "false" ]; then
+  export CORS_ALLOW_PRIVATE_NETWORK_ACCESS=""  
+fi
+
 # This is the primary logic to determine the s3 host used for the
 # upstream (the actual proxying action) as well as the `Host` header
 #
diff --git a/standalone_ubuntu_oss_install.sh b/standalone_ubuntu_oss_install.sh
index dbeb3f1..ea74e7f 100644
--- a/standalone_ubuntu_oss_install.sh
+++ b/standalone_ubuntu_oss_install.sh
@@ -30,6 +30,17 @@ fi
 
 failed=0
 
+parseBoolean() {
+  case "$1" in
+    TRUE | true | True | YES | Yes | 1)
+      echo 1
+      ;;
+    *)
+      echo 0
+      ;;
+  esac
+}
+
 required=("S3_BUCKET_NAME" "S3_SERVER" "S3_SERVER_PORT" "S3_SERVER_PROTO"
 "S3_REGION" "S3_STYLE" "ALLOW_DIRECTORY_LIST" "AWS_SIGS_VERSION")
 
@@ -142,6 +153,8 @@ if [ "${to_install}" != "" ]; then
   systemctl stop nginx
 fi
 
+cors_enabled="$(parseBoolean "${CORS_ENABLED}")"
+
 echo "▶ Adding environment variables to NGINX configuration file: /etc/nginx/environment"
 cat > "/etc/nginx/environment" << EOF
 # Enables or disables directory listing for the S3 Gateway (true=enabled, false=disabled)
@@ -179,18 +192,24 @@ PROXY_CACHE_VALID_NOTFOUND=${PROXY_CACHE_VALID_NOTFOUND:-'1m'}
 # Proxy caching time for response code 403
 PROXY_CACHE_VALID_FORBIDDEN=${PROXY_CACHE_VALID_FORBIDDEN:-'30s'}
 # Enables or disables CORS for the S3 Gateway (true=enabled, false=disabled)
-CORS_ENABLED=${CORS_ENABLED:-'false'}
+CORS_ENABLED=${cors_enabled:-'false'}
 # Configure portion of URL to be removed (optional)
 STRIP_LEADING_DIRECTORY_PATH=${STRIP_LEADING_DIRECTORY_PATH:-''}
 # Configure portion of URL to be added to the beginning of the requested path (optional)
 PREFIX_LEADING_DIRECTORY_PATH=${PREFIX_LEADING_DIRECTORY_PATH:-''}
 EOF
 
+# Normalize the CORS_ENABLED environment variable to a numeric value
+# so that it can be easily parsed in the nginx configuration.
+cat >> "/etc/nginx/environment" << EOF
+CORS_ENABLED="${cors_enabled}"
+EOF
+
 # By enabling CORS, we also need to enable the OPTIONS method which
 # is not normally used as part of the gateway. The following variable
 # defines the set of acceptable headers.
 set +o nounset   # don't abort on unbound variable
-if [ "${CORS_ENABLED}" == "1" ]; then
+if [ "${cors_enabled}" == "1" ]; then
     cat >> "/etc/nginx/environment" << EOF
 LIMIT_METHODS_TO="GET HEAD OPTIONS"
 LIMIT_METHODS_TO_CSV="GET, HEAD, OPTIONS"
@@ -202,6 +221,18 @@ LIMIT_METHODS_TO_CSV="GET, HEAD"
 EOF
 fi
 
+# See documentation for this feature. We do not parse this as a boolean
+# since "true" and "false" are the required values of the header this populates
+if [ "${CORS_ALLOW_PRIVATE_NETWORK_ACCESS}" != "true" ] && [ "${CORS_ALLOW_PRIVATE_NETWORK_ACCESS}" != "false" ]; then
+    cat >> "/etc/nginx/environment" << EOF
+CORS_ALLOW_PRIVATE_NETWORK_ACCESS=""
+EOF
+else
+    cat >> "/etc/nginx/environment" << EOF
+CORS_ALLOW_PRIVATE_NETWORK_ACCESS="${CORS_ALLOW_PRIVATE_NETWORK_ACCESS}"
+EOF
+fi
+
 # This is the primary logic to determine the s3 host used for the
 # upstream (the actual proxying action) as well as the `Host` header
 #

``` 
(the above diff also contains a small fix for a bug I discovered in the VM bootstrap script regarding the `CORS_ENABLED` variable handling)

This would mean the changes to `cors.conf.template` would then just be this:
```
add_header 'Access-Control-Allow-Private-Network' '${CORS_ALLOW_PRIVATE_NETWORK_ACCESS}';
```

Sorry about the churn on this PR but I'm also learning as I touch various parts of this project.

4141done:
Oh- and don't worry about any changes in the Ubuntu standalone script.  That one is a bit more tricky to verify so I'll handle.adding that after this pr merges

danielcompton:
Thanks for the guidance @4141done. I've updated the branch with your suggestions. I'm also happy for you to make any other changes directly if that's easier for you.

4141done:
Hey apologies for my slowness - just wanted to let you know that I've received this and it looks good - I'm just adding this to the VM startup script and running a quick test then I think we can merge soon.

4141done (approved):
Thank you for your work and patience with us on this!

danielcompton:
Amazing, thanks so much! I appreciate your help on this.

Issue #192, opened 2023-12-19
Check environment fails with IMDSv2


peter-svensson:
During startup a check is performed that the necessary `env` variables are present in the container.
If using in K8s running on EC2 instances with IMDSv2 enabled/enforced the checks will fail:

```shell
/docker-entrypoint.sh: Launching /docker-entrypoint.d/00-check-for-required-env.sh
Required S3_ACCESS_KEY_ID environment variable missing
Required S3_SECRET_KEY environment variable missing
```

I guess the reason for this are the following lines:
https://github.com/nginxinc/nginx-s3-gateway/blob/bb03e8889025b76e0af51f40882ca67672d18d28/common/docker-entrypoint.d/00-check-for-required-env.sh#L47-L48

Calling the `http://169.254.169.254` endpoint requires a token (which is correctly done here for example): https://github.com/nginxinc/nginx-s3-gateway/blob/bb03e8889025b76e0af51f40882ca67672d18d28/common/etc/nginx/include/awscredentials.js#L345-L347 
I guess we need to update `00-check-for-required-env.sh`  to fetch the token first as well, like:

```shell
elif TOKEN=`curl -X PUT --silent --fail --connect-timeout 2 --max-time 2 "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metada
ta-token-ttl-seconds: 21600"` && curl  -H "X-aws-ec2-metadata-token: $TOKEN" --output /dev/null --silent --head --fail --connect-timeout
2 --max-time 5 "http://169.254.169.254"; then 
```

Reference: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#instance-metadata-returns

PR #193, opened 2023-12-19
fix: handle IMDSv2 auth when checking if we are running on EC2 instance

peter-svensson:
Fix for #192


4141done:
Hi @peter-svensson - thank you for your contribution and detailed bug report.  This change looks good to me in terms of handling IMDSv2.  I was initially worried that it would break IMDSv1 users but as you pointed out in the issue we use this path in a similar logic branch in the Javascript code itself. 

Merging with gratitude 💚

PR #194, opened 2023-12-19
[WIP] Move Integration Tests to NodeJS

4141done:
> [!NOTE]  
> This PR is a work in progress but feedback on direction is welcome.

## What
This PR moves the existing integration test suite into a Jest + Superagent-based suite.
The goal is to mostly replicate the behavior of the current test suite while also accounting for the large number of potential configurations for the project.

### Organization
1. A basic test set handling encoding, basic usage, etc.
2. Use-case based tests handling (subject to change)
    1. File/directory browsing
    2. Caching specific
    3. Single page application/asset server
    4. object path acrobatics

### Design
The basic execution flow for each test will be:
1. Ensure a clean state in Minio
1. Create a bucket using the Minio Client
2. Write the required test files to the bucket
3. Build and start an s3-gateway container using the supplied settings for the scenario
4. Use supertest to send and verify requests against the running container.
5. Clean up the Minio bucket.

The principles we will try to follow are:
* Test file structure specific to each use-case based test suite
* Starting a general transition to thinking about use cases rather than individual config (although that will always be an option)
* Decoupling general container build from integration test run


ciroque (commented):
Definitely a fan of moving away from shell-based test suites. This seems logical and approachable. Curious what the Contributor Journey will be. 

Do new tests go in the `__tests__/basic.test.js` file, or new files based on the functionality being tests? I'm guessing new files. If that's correct, will there he a template or generator for creating new test files?

Overall this looks good. I have not written Jest / JavaScript tests in a long time so details are new again for me. Regardless, this is a good start!

tsconfig.json
ciroque, line 1: For me all these commented lines are quite distracting. Though I cannot imagine I'd be in this file very often.

ciroque (commented):
LGTM.

dekobon:
I would do an explicit call out for the need to install node and/or npm modules.

Issue #196, opened 2024-01-05
Arm64 build


maxemann96:
**Is your feature request related to a problem? Please describe.**
I want to run a kubernetes cluster with only arm64 nodes, as a result I need an arm64 build for every software I use.

**Describe the solution you'd like**
Also building a docker image for arm64

**Describe alternatives you've considered**
No alternatives

**Additional context**
The nginx base image is already available as arm64

Thanks in advance for creating and maintaining this image. Very useful for my use case.

4141done:
Hello!

Thank you for bringing this up! It is on my list of things to do but having someone complain certainly helps with prioritization.  You should see this change coming in fairly soon.  See my reply to [this issue](https://github.com/nginxinc/nginx-s3-gateway/issues/195#issuecomment-1879106854) regarding a refactor I'm doing at the moment to decouple the integration tests from the build process and move the integration test suite into Javascript.  I'm tracking that work here: https://github.com/nginxinc/nginx-s3-gateway/issues/191

hmehta:
Hi!

This would be so wonderful! We are moving from AMD64->ARM64 and this is one of the few things we are using that not yet supports ARM64. So a big upvote from here for bumping this up a notch on the todo list :)

ehginanjar:
Hi @maxemann96 

I've had similar case, building the image so I can test it on my M1 Pro. In this case I used docker, so I tried to build the image by using this command `docker buildx build --platform linux/arm64`. So maybe this solution can be tested on your side as well. Cheers!

4141done:
@maxemann96 and @hmehta :  I have a provisional PR for an overhaul of the build and test pipeline in case you have comments here: https://github.com/nginxinc/nginx-s3-gateway/pull/197
You don't necessarily have to review it but any comments are welcome.

More importantly, it produced a test image that you can see here: https://github.com/nginxinc/nginx-s3-gateway/pkgs/container/nginx-s3-gateway%2Fnginx-oss-s3-gateway/166168166?tag=latest-new-build-test

Would you mind testing to see if this works for you?

maxemann96:
@4141done the test image is running properly, thanks :)

hmehta:
@4141done I confirm that the image works for us too. Thanks a lot!

4141done:
Thank you both for confirming.  Arm64 images will start being built after that PR is merged.  I'm just waiting on code review from my co-maintainers.

Issue #191, opened 2023-12-15
Translate integration test suite to Javacript


4141done:
**Is your feature request related to a problem? Please describe.**
Currently the integration test suite has the following issues:
* It is not separate from the image build
* It is tough to understand at first. Especially for new contributors
* New configuration options combined with bash's inflexibility are leading to an increasingly difficult to maintain test suite.

**Describe the solution you'd like**
1. Decouple image build from integration test run (this will help us clean up the build task and start building multi-arch images)
2. Use a NodeJS-based test runner to perform integration tests

**Describe alternatives you've considered**
I could be possible to refactor the bash-based runner to handle things better but:
* It seems contributors are more familiar with Javascript than NGINX/Bash. This will lower dependence on the maintainer for tests.
* The new maintainer is more comfortable in JS.

dekobon:
Closing as superseded. The plan of record for test-suite maintainability changed: rather than rewriting the integration suite in JavaScript, the project is migrating to a make-native workflow — #536 (e5e3904) replaced the GNUmakefile with a comprehensive build/test/lint target set as the only supported interface, and `test.sh` is being burned down into make-native pieces. The underlying goals here (decoupling build from test, easier maintenance and extension) are being addressed on that path instead.

PR #197, opened 2024-01-09
Build multi-architecture images

4141done:
# What
Allows us to push multi-architecture builds for `linux/amd64` and `linux/arm64` on merge to master. Fixes: https://github.com/nginxinc/nginx-s3-gateway/issues/196

The flow looks like this:

```
                                   ┌──────────────────┐        ┌────────────────┐      ┌────────────────┐
    ┌─────────┐     ┌─────────┬────► Build Latest NJS ├────────►Test Latest NJS ├─────►│Push Latest NJS │
    │Build OSS├────►│Test OSS │    └──────────────────┘        └────────────────┘      └────────────────┘
    └─────────┘     └──┬──────┤
                       │      │    ┌──────────────────┐       ┌──────────────────┐     ┌─────────────────┐
                       │      └────►Build Unprivileged├───────►Test Unprivileged ├────►│Push Unprivileged│
                       │           └──────────────────┘       └──────────────────┘     ├────────┬────────┘
                       │                                                               ├────────┤
                       └──────────────────────────────────────────────────────────────►│Push OSS│
                                                                                       └────────┘


```
Tests are performed by building the relevant image and tagging it with the image name the tests expect.  This follows the original implementation in `test.sh`.  This will likely change to an explicit image name specification as we move the test suite in to Javascript but it's being maintained for now in the interest of simplicity.

## Goals
* To decouple image build from the test run. This will help as we refactor the test suite for this issue https://github.com/nginxinc/nginx-s3-gateway/issues/191.
* Not have CI take a long time due to multi-architecture builds for interim pushes to pull requests.
* CI should fail fast if any of the image builds or tests fail
* Parallelize builds as much as possible.

## Notes
* Since the `unprivileged` and `latest-njs` images build off the base `oss` image, there were some strange moves that had to be done to make that work without rebuilding the base image every time. Specifically, we had to set `setup-buildx-action` to run in `docker` driver mode so we can simply `load` the base image. Otherwise we would have had to use a local repository.  However, in `docker` mode `upload-artifact` doesn't like the file produced so we have to save the file again.
* I wanted to build images once and then run tests against them and push at the end. However, there was not a clean way to get the full multi-architecture images all the way to the push step so I just build them again against all architectures and perform tests in the runner architecture.  This saves us from needing conditionals in the test portions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -172,132 +172,76 @@ jobs:
       - name: Run tests - latest njs version
         run: ./test.sh --latest-njs --type oss
 
+  build-unprivileged-for-test:
+    runs-on: ubuntu-latest
+    needs: test-oss
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker
+      - name: Download artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: oss
+          path: /tmp
+      - name: Load image
+        run: |
+          docker load --input /tmp/oss.tar
+      - name: Build and load oss image
+        uses: docker/build-push-action@v5
+        with:
+          file: Dockerfile.unprivileged
+          context: .
+          tags: nginx-s3-gateway:unprivileged-oss
+          load: true
+      # Save manually here since we need to use the `docker` buildx `driver` but that can't output
+      # a file that upload-artifact likes.
+      - name: save image
+        run: |
+          docker save nginx-s3-gateway:unprivileged-oss > /tmp/unprivileged.tar
+      - name: Upload artifact - unprivileged
+        uses: actions/upload-artifact@v3
+        with:
+          name: unprivileged
+          path: /tmp/unprivileged.tar
 
-  #     - name: Build and export - unprivileged
-  #       uses: docker/build-push-action@v5
-  #       with:
-  #         file: Dockerfile.unprivileged
-  #         context: .
-  #         tags: localhost:5000/nginx-s3-gateway:unprivileged
-  #         push: false
-  #         outputs: type=oci,dest=/tmp/unprivileged.tar
-  #         platforms: linux/amd64,linux/arm64
-  #         build-contexts: |
-  #           nginx-s3-gateway=docker-image://localhost:5000/nginx-s3-gateway:oss
-  #     - name: Upload artifact - unprivileged
-  #       uses: actions/upload-artifact@v3
-  #       with:
-  #         name: unprivileged
-  #         path: /tmp/unprivileged.tar
-  # test-oss:
-  #   runs-on: ubuntu-latest
-  #   if: github.ref != 'refs/heads/master'
-  #   needs: build
-  #   steps:
-  #     - uses: actions/checkout@v4
-  #     ## Start tests.  Keep things here because we can't keep the registry between jobs
-  #     ## TODO: Try saving the artifact after all are built?  What happens to the other archs?
-  #     - name: Install dependencies
-  #       run: sudo apt-get update -qq && sudo apt-get install -y curl wait-for-it
-  #     - name: Restore cached binaries
-  #       id: cache-binaries-restore
-  #       uses: actions/cache/restore@v3
-  #       with:
-  #         path: .bin
-  #         key: ${{ runner.os }}-binaries
-  #     - name: Install MinIO Client
-  #       run: |
-  #         mkdir .bin || exit 0
-  #         cd .bin
-  #         curl --insecure --retry 6 --fail --location --output mc.RELEASE.2023-06-19T19-31-19Z "https://dl.min.io/client/mc/release/linux-$(dpkg --print-architecture)/archive/mc.RELEASE.2023-06-19T19-31-19Z"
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -8,16 +8,53 @@ on:
   pull_request:
     branches: [ master ]
 
-# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+
+env:
+  CI: true
```
4141done, line 13: **annotation**: This is temporary until we get the whole test suite refactored. For now we skip the image build step in the test script since this pipeline ensures that the correct images are available.

.github/workflows/main.yml
```diff
@@ -8,16 +8,53 @@ on:
...
+        with:
+          file: Dockerfile.oss
+          context: .
+          tags: nginx-s3-gateway , nginx-s3-gateway:oss
+          outputs: type=docker,dest=/tmp/oss.tar
+      - name: Upload artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: oss
+          path: /tmp/oss.tar
```
4141done, line 50: **annotation**: The other two images use this as a base.  Therefore we export the built image for use in:
* The tests for this image
* The build process for the `latest-njs` and `unprivileged` images

.github/workflows/main.yml
```diff
@@ -75,79 +134,159 @@ jobs:
...
+
+  build-unprivileged-for-test:
+    runs-on: ubuntu-latest
+    needs: test-oss
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker
```
4141done, line 161: **annotation:**
If we use the `docker-container` driver that buildx uses by default, a `docker load` does not get the image in a context where the build process can access it.  In order to do this you must use a local registry as detailed in this guide: https://docs.docker.com/build/ci/github-actions/named-contexts/#using-with-a-container-builder

.github/workflows/main.yml
```diff
@@ -75,79 +134,159 @@ jobs:
...
+          file: Dockerfile.oss
+          context: .
+          push: true
+          platforms: linux/amd64,linux/arm64
+          provenance: false
+          tags: |
+            ghcr.io/${{ github.repository }}/nginx-oss-s3-gateway:latest-${{ steps.date.outputs.date }}
+            ghcr.io/${{ github.repository }}/nginx-oss-s3-gateway:latest
+            nginxinc/nginx-s3-gateway:latest-${{ steps.date.outputs.date }}
+            nginxinc/nginx-s3-gateway:latest
```
4141done, line 264: **annotation**:
This is nice, the plugin lets you log in to both registries then pushes tot the correct one based on the tag.

.github/workflows/main.yml
```diff
@@ -75,79 +134,159 @@ jobs:
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push image [oss]
+        uses: docker/build-push-action@v5
+        with:
+          file: Dockerfile.oss
+          context: .
+          push: true
+          platforms: linux/amd64,linux/arm64
+          provenance: false
```
4141done, line 259: **annotation**:
This is a little workaround for the github packages repo.  The `provenance` file is a docker thing not an OCI thing so you'll get an `unknown/unknown` architecture listed in gh packages if you don't set this.

ref: https://docs.docker.com/build/attestations/slsa-provenance/

If we want these to go to dockerhub, we could also separate the builds again but I think the simplicity of having them both push is more desireable.

.github/workflows/main.yml
```diff
@@ -75,79 +134,159 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker
+      - name: Download artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: oss
+          path: /tmp
```
dekobon, line 166: I would check to see if github actions has defined an environment variable for the temp path. Generally, it isn't great to assume that `/tmp` is the system temp directory. Yes, I know it is in this case and this is a nit.

.github/workflows/main.yml
```diff
@@ -75,79 +134,159 @@ jobs:
...
+      # Save manually here since we need to use the `docker` buildx `driver` but that can't output
+      # a file that upload-artifact likes.
+      - name: save image
+        run: |
+          docker save nginx-s3-gateway:unprivileged-oss > /tmp/unprivileged.tar
+      - name: Upload artifact - unprivileged
+        uses: actions/upload-artifact@v3
+        with:
+          name: unprivileged
+          path: /tmp/unprivileged.tar
```
dekobon, line 186: Refresh my memory if the uploaded artifacts eventually get cleaned up. I have a vague memory that they do.

dekobon (approved):
Looks good. I'm looking forward to seeing this merged.

.github/workflows/main.yml
4141done: There's a default setting for the repo which I believe is 30 days.  However based on this comment I'm going to update the retention explicitly since so it's obvious from the script. 

`retention-days: 1` One day is the minimum which is really too long for this but should serve us fine.

Also worth noting that artifacts are [only accessible from the workflow in which they were created](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#downloading-artifacts-during-a-workflow-run) so there's no chance of us pulling an old artifact which I was worried about initially.

.github/workflows/main.yml
4141done: Thank you for this note.  I discovered `${{ runner.temp }}` which serves the purpose well.

Issue #206, opened 2024-01-30
using standalone_ubuntu_oss_install.sh on ec2 fail the EC@ instance check


orensharoni:
**The bug**
when running the standalone_ubuntu_oss_install.sh on an ec2 instance the check fail and requires a 
AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY despite running on an ec2 with the proper Role

**To Reproduce**
1. clone to an ec2 instance
2. run standalone_ubuntu_oss_install.sh
3. See error

**Expected behavior**
the script needs to pass the ec2 test.

**Your environment**
* Version of the repo - master
* 
* S3 backend implementation you are using -AWS
* Deploying stand-alone using the script
* NGINX OSS
* Authentication method-IAM

**Additional context**
I think it is since the check `curl --output /dev/null --silent --head --fail --connect-timeout 2 "http://169.254.169.254"`
refers to IMDSv1 and it needs to change to IMDSv2, i used the following:
`TOKEN=`curl  --silent  --fail --connect-timeout 2 -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600"` \
    && curl --output /dev/null --silent --head --fail --connect-timeout 2 -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/`

4141done:
Thank you for your report @orensharoni , I'll take a look this week.  We did have a PR come in that addressed this in the container startup script so it's likely we just need to port that change over to the script you're using.  https://github.com/nginxinc/nginx-s3-gateway/pull/193

Feel free to open a PR if you get to it before I do 🐳

orensharoni:
hi,
will love to help , i have the fix working locally I can't push a branch to create the PR - no contributor privlige

4141done:
> hi, will love to help , i have the fix working locally I can't push a branch to create the PR - no contributor privlige

That would be great! You'll need to first fork this repository, create a branch on your local fork, then follow this guide to open a pull request: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork


Question for you - It seems that most new ec2 instances have support for both IMDSv2 and IMDSv1.  Would it make sense to leave the `elif` branch for IMDSv1 as a fallback if someone has an old instance?  For example:

```bash
[...]

PR #212, opened 2024-02-12
Issue #206 -ec2 check to support IMDSv1 and IMDSv2 + moving env vars to conf file

orensharoni:
we added support of IMDSv2 to the EC2 check with a fall back to IMDSv1,
also, some nginx env declarations were moved from the environment file to the nginx.conf file

### DIFF
--- a/standalone_ubuntu_oss_install.sh
+++ b/standalone_ubuntu_oss_install.sh
@@ -36,8 +36,12 @@ required=("S3_BUCKET_NAME" "S3_SERVER" "S3_SERVER_PORT" "S3_SERVER_PROTO"
 if [ ! -z ${AWS_CONTAINER_CREDENTIALS_RELATIVE_URI+x} ]; then
   echo "Running inside an ECS task, using container credentials"
   uses_iam_creds=1
+elif TOKEN=$(curl -X PUT --silent --fail --connect-timeout 2 --max-time 2 "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 21600") && \
+  curl -H "X-aws-ec2-metadata-token: $TOKEN" --output /dev/null --silent --head --fail --connect-timeout 2 --max-time 5 "http://169.254.169.254"; then 
+  echo "Running inside an EC2 instance, using IMDSv2 for credentials"
+  uses_iam_creds=1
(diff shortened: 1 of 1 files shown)


4141done:
Thank you for your contribution, @orensharoni  - have been able to verify that this fixes your issue from https://github.com/nginxinc/nginx-s3-gateway/issues/206 ?

I'll stand up an ec2 env and give it a check too.

4141done:
> yes, I used it on my dev setup and the ec2 test and deployment work, the access_key and access_token should be also verified if posible

Excellent, thank you for confirming.  I'll do a quick check on my end and then merge

standalone_ubuntu_oss_install.sh
```diff
@@ -312,12 +316,12 @@ EOF
 # Only include these env vars if we are not using a instance profile credential
 # to obtain S3 permissions.
 if [ $uses_iam_creds -eq 0 ]; then
-  cat >> "/etc/nginx/environment" << EOF
+  cat >> "/etc/nginx/nginix.conf" << EOF
```
4141done, line 319:
```suggestion
  cat >> "/etc/nginx/nginx.conf" << EOF
```

standalone_ubuntu_oss_install.sh
```diff
@@ -312,12 +316,12 @@ EOF
...
 env AWS_ACCESS_KEY_ID;
 env AWS_SECRET_ACCESS_KEY;
 EOF
   if [[ -v AWS_SESSION_TOKEN ]]; then
-    cat >> "/etc/nginx/environment" << EOF
+    cat >> "/etc/nginx/nginix.conf" << EOF
```
4141done, line 324:
```suggestion
    cat >> "/etc/nginx/nginx.conf" << EOF
```

4141done (changes requested):
Thanks again for your contribution.  I tested these changes with the token approach and was able to verify it works well.

If you can correct the two typoes I highlighted we can go ahead and merge.

standalone_ubuntu_oss_install.sh
orensharoni: done

standalone_ubuntu_oss_install.sh
orensharoni: done

4141done (approved):
Thank you again for your contribution! Also special thanks to @HighOnMikey
for noticing the wrong file path in the env var inclusion portion

PR #229, opened 2024-04-18
docs: include instructions for usage with AWS S3 One Zone

4141done:
# What
Instructions to use AWS S3 One Zone. For a further explanation, please see the docs files in the PR.


deployments/s3_express/settings.s3express.example
```diff
@@ -0,0 +1,21 @@
+S3_BUCKET_NAME=bucket_name=my-bucket-name--usw2-az1--x-s3
```
hveiga, line 1: I think there's a typo here: `bucket_name=` should not be present in the value.

common/etc/nginx/include/s3gateway.js
```diff
@@ -86,7 +86,7 @@ const INDEX_PAGE = "index.html";
  * Constant defining the service requests are being signed for.
  * @type {string}
  */
-const SERVICE = 's3';
+const SERVICE = 's3express';
```
hveiga, line 89: I believe we can undo this change.

4141done:
Now that we've verified that this change works, I'm going to follow @hveiga 's original direction and add the `S3_SERVICE` configuration variable which will default to `s3` and may be one of `s3express` or `s3`.

Setting this variable to `s3express` will have the following effects:
1. Change the "service" used to sign the requests with the V4 header to `s3express`. Currently the gateway works without this step, but it's advised in the documentation [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-express-security-best-practices.html).
2. Set the upstreams for both open source and plus to prepend the bucket name to the server address as seen in this PR

I may do a bit of a refactor to accomplish this since we currently have a good number of places where the host is being assembled:

1. https://github.com/nginxinc/nginx-s3-gateway/blob/7f3064b2c5cf08871164609de8d7e3f972e0a6a5/common/etc/nginx/templates/default.conf.template#L22 for the `Host` header
1. https://github.com/nginxinc/nginx-s3-gateway/blob/7f3064b2c5cf08871164609de8d7e3f972e0a6a5/common/etc/nginx/include/s3gateway.js#L230 for the host used to sign the v4 headers
1. https://github.com/nginxinc/nginx-s3-gateway/blob/7f3064b2c5cf08871164609de8d7e3f972e0a6a5/oss/etc/nginx/templates/upstreams.conf.template#L10 for the actual proxying of the server.

Based on my [reading](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html), for a `virtual` style request signed with the V4 signature we want:

* Upstream pointing to: `https://bucket-name.s3.region-code.amazonaws.com` (`S3_BUCKET_NAME` and ${S3_SERVER}` in our configuration model)
* `Host` header set to same
*  The `host` portion of the AWS v4 signature should match the `Host` header ([ref](https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html))

This is what is happening at the moment but for forward maintainability I'd like to see if we can colocate the logic as much as possible.  It may not be possible to avoid duplication due to the fact that the `upstream` blocks need to have the server address templated in before the nginx server is started.

hveiga (approved):
LGTM

4141done:
Thank you for the approval @hveiga .  I am going to have some other NGINX folks look over this before I merge.  I've also opened a discussion here https://github.com/nginxinc/nginx-s3-gateway/discussions/231 on the chance that this change could cause issues for others.  I will wait for a while to give people a chance to speak up before merging. I will make sure that you are credited for this contribution when the PR is merged

common/etc/nginx/templates/default.conf.template
```diff
@@ -19,11 +19,10 @@ map $uri_full_path $uri_path {
...
 
-map $S3_STYLE $s3_host_hdr {
-    virtual "${S3_BUCKET_NAME}.${S3_SERVER}";
-    path    "${S3_SERVER}:${S3_SERVER_PORT}";
-    default "${S3_BUCKET_NAME}.${S3_SERVER}";
-}
+# FINAL_S3_SERVER is set in the startup script
+# (either ./common/docker-entrypoint.sh or ./standalone_ubuntu_oss_install.sh)
+# based on the S3_STYLE configuration option.
+js_var $s3_host ${FINAL_S3_SERVER};
```
4141done, line 25: **Annotation:**
This is the primary change.  We determine this value in the startup script then template it in before NGINX starts.  While before this value was calculated in three different places, using `js_var` allows access to its value both in the njs script and in the configuration.

.github/workflows/main.yml
```diff
@@ -54,6 +54,9 @@ jobs:
   test-oss:
     runs-on: ubuntu-22.04
     needs: build-oss-for-test
+    strategy:
+      matrix:
+        path_style: [virtual, virtual-v2]
```
4141done, line 59: **Annotation:**
We are only testing both path styles at the main test as a compromise on test time.  This is the primary place where something might be different and the rest of tests are variations on deployment options.

4141done:
Thanks again for all your help on this change @hveiga  - please note that in the end I wound up adding the `S3_STYLE` config option `virtual-v2` to cut down on risk so you'll need to use that config.  Please let me know if you experience any issues and we can fix forward.

hveiga:
> Thanks again for all your help on this change @hveiga - please note that in the end I wound up adding the `S3_STYLE` config option `virtual-v2` to cut down on risk so you'll need to use that config. Please let me know if you experience any issues and we can fix forward.

Thanks to you for the final push to get this in. I'll be testing today and I can report back if I find any issues. 🙌

hveiga:
I just tested both S3 General and S3 Express buckets with `ghcr.io/nginxinc/nginx-s3-gateway/nginx-oss-s3-gateway:latest-20240425` and it works as expected 🎉 🙌

4141done:
> I just tested both S3 General and S3 Express buckets with `ghcr.io/nginxinc/nginx-s3-gateway/nginx-oss-s3-gateway:latest-20240425` and it works as expected 🎉 🙌

Woohoo! I'm going to close out the issue and the discussion point then.

PR #230, opened 2024-04-19
Fix issue where bucket name is appended twice on path-style index page requests

### LINKED ISSUE #210: Path style breaks fetching INDEX_PAGE
[`loadContent`](https://github.com/nginxinc/nginx-s3-gateway/blob/c687663/common/etc/nginx/include/s3gateway.js#L390) will always attempt fetching the `INDEX_PAGE` by building the URI path using [`s3uri()`](https://github.com/nginxinc/nginx-s3-gateway/blob/c687663/common/etc/nginx/include/s3gateway.js#L283).

When using `S3_STYLE = 'path'`, this will prepend the URI path with the S3 bucket. However, the fetch is made against nginx itself, meaning the URI path will always be incorrect for the requested index. 

**To Reproduce**

Set the following variables:

```
ALLOW_DIRECTORY_LIST=true
PROVIDE_INDEX_PAGE=true
S3_STYLE=path
```

**Expected behavior**

Incorrect URI currently being built: `http://127.0.0.1/<bucket_name>/<request_path>/<INDEX_PAGE>`

Expected URI: `http://127.0.0.1/<request_path>/<INDEX_PAGE>`

**Environments Tested**

* `master` @ `c687663`
* `ghcr.io/nginxinc/nginx-s3-gateway/nginx-oss-s3-gateway:latest`
* Backends: AWS, Ceph, OpenStack Swift

**Workaround**

I would submit a PR but I'm not sure which direction would be preferred by the project to go about building the URI path. However, here's a quick fix to demonstrate the issue:

From

```js
const uri = s3uri(r);
```

To

```js
let uri;
if (S3_STYLE === 'path') {
    uri = `${r.uri}${INDEX_PAGE}`;
} else {
    uri = s3uri(r);
}
```

4141done:
# What
Aims to fix https://github.com/nginxinc/nginx-s3-gateway/issues/210

The gateway has a feature that allows serving of the base index page that lists files in the directory if no `index.html` exists in the requested folder. In order to do this it performs a test request against `<path>/ index.html` to check for an existing index page.

For path-style URIs where the bucket name must be provided, this resulted in the bucket name being prepended to the request twice.

For more details on the presentation and required solution, please see the issue.

## Fix
I wanted to keep `s3uri` in charge of the behavior here, so chose to add an options parameter to it with the option `preserveBasePath` so that we can ask it to do its other work, but give back a path without anything prepended. In the future this object could be used to offer control over other operations


common/etc/nginx/include/s3gateway.js
```diff
@@ -278,30 +278,42 @@ function s3BaseUri(r) {
...
         // This is a path that will resolve to an index page
         if (PROVIDE_INDEX_PAGE  && _isDirectory(uriPath) ) {
             uriPath += INDEX_PAGE;
         }
-        path = _escapeURIPath(basePath + uriPath);
+        path = _escapeURIPath(`${basePath}${uriPath}`);
     }
 
-    utils.debug_log(r, 'S3 Request URI: ' + r.method + ' ' + path);
+    utils.debug_log(r, `S3 Request URI: ${r.method} ${path}`);
```
4141done, line 316: **Annotation**
These changes are not part of the main change - just some visual cleanup

dekobon:
Superseded by #559

PR #252, opened 2024-05-15
add eks pod identity credentials support

tieum:
### Proposed changes

Add support of [EKS pod identities](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html)

### Checklist

- [ ] I have signed the [F5 Contributor License Agreement (CLA)](https://github.com/f5/.github/blob/main/CLA/cla-markdown.md).
- [ ] If applicable, I have added tests that prove my fix is effective or that my feature works.
   --> I didn't manage to add the correct test, however I didn't find any for webIdentity auth neither 

- [ ] I have updated any relevant documentation (e.g. [`README.md`](/README.md)).
  --> there is a [nice .png](https://github.com/nginxinc/nginx-s3-gateway/blob/main/docs/img/nginx-s3-gateway-signature-flow.png) for which I didn't have the source to add `_fetchEKSPodIdentityCredentials` in the list

### DIFF
--- a/common/etc/nginx/include/awscredentials.js
+++ b/common/etc/nginx/include/awscredentials.js
@@ -59,6 +59,12 @@ const EC2_IMDS_TOKEN_ENDPOINT = 'http://169.254.169.254/latest/api/token';
 const EC2_IMDS_SECURITY_CREDENTIALS_ENDPOINT = 'http://169.254.169.254/latest/meta-data/iam/security-credentials/';
 
 /**
+ * URL to EKS Pod Identity Agent credentials endpoint
+ * @type {string}
+ */
+const EKS_POD_IDENTITY_AGENT_CREDENTIALS_ENDPOINT = 'http://169.254.170.23/v1/credentials'
+
+/**
  * Offset to the expiration of credentials, when they should be considered expired and refreshed. The maximum
  * time here can be 5 minutes, the IMDS and ECS credentials endpoint will make sure that each returned set of credentials
  * is valid for at least another 5 minutes.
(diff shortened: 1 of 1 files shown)


tieum:

4141done:
Thank you for your contribution, @tieum 🎉 I will make some time to understand and review this week.  
Based on my quick look, can you confirm some things for me?
1. This change will require the addition of a new config variable `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE`?
1. Can you tell me a bit more about your use case?  I have not tried this in eks and I'm curious to understand how widely applicable this might be

tieum:
👋 @4141done 

1. The `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` environnement variable is available in the container [when using EKS Pod Identity](https://docs.aws.amazon.com/eks/latest/userguide/pod-id-how-it-works.html#pod-id-agent-pod), the same way than `AWS_WEB_IDENTITY_TOKEN_FILE` is available when using IRSA
I've updated the `getting_started.md` on how to set it up

2.  EKS Pod Identity was released in December 2023 and provides an alternative way to grant workload access to AWS, the differences with IAM Roles for Services Accounts can be found [here](https://docs.aws.amazon.com/eks/latest/userguide/service-accounts.html#service-accounts-iam), a Datadog Security Labs [blog post ](https://securitylabs.datadoghq.com/articles/eks-pod-identity-deep-dive/#main-differences-between-irsa-and-eks-pod-identity) also sums it up nicely

4141done:
Thank you for the explanation and for the very clean pull request.  I think this is a great addition to the library. I have two requests before we merge:

1. Can you try to add a test for this?  There's an example to follow [here](https://github.com/nginxinc/nginx-s3-gateway/blob/1a1b429493964ba634199334bd5889ee2201277a/test/unit/awscredentials_test.js#L181). I know that our test suite isn't very standard since it's run using the njs binary so let me know if you have questions or would like me to jump in.
2. Can you take a look at [this file](https://github.com/nginxinc/nginx-s3-gateway/blob/main/standalone_ubuntu_oss_install.sh) and see if we need to look for the file in the VM case too? I'm guessing not since this is EKS but I'd like to get your more informed take on whether it could be necessary in some situations.

tieum:
1. I added a test let me know what you think
2. I checked the standalone install script and you are right, in the VM case this auth mechanism won't be available / there is no need to check for it

4141done (approved):
👍 🎉 🐳 
This all looks great.  Once again we very much appreciate your contribution!

PR #274, opened 2024-06-26
fix: update NGINX Plus Dockerfile

alessfg:
### Proposed changes

This PR refactors the NGINX Plus Dockerfile by using the NGINX Plus Dockerfile found here [gist.github.com/nginx-gists/36e97fc87efb5cf0039978c8e41a34b5](https://gist.github.com/nginx-gists/36e97fc87efb5cf0039978c8e41a34b5) as a starting point.

This PR also refactors the various ENV instructions in both the NGINX OSS and NGINX Plus Dockerfiles to use the new guidelines where the key=value pair is separated by "=" instead of a whitespace.

### Checklist

Before creating a PR, run through this checklist and mark each as complete:

- [x] I have signed the [F5 Contributor License Agreement (CLA)](https://github.com/f5/.github/blob/main/CLA/cla-markdown.md).


Dockerfile.buildkit.plus
```diff
@@ -1,83 +1,102 @@
...
-    ln -sf /dev/stderr /var/log/nginx/error.log; \
-    chmod -R -v +x /docker-entrypoint.sh /docker-entrypoint.d/*.sh
+# We create an NGINX Plus image based on the official NGINX Plus Dockerfiles (https://gist.github.com/nginx-gists/36e97fc87efb5cf0039978c8e41a34b5) and modify it by:
+# 1. Explicitly installing the version of njs coded in the environment variable above.
+# 2. Adding configuration files needed for proxying private S3 buckets.
+# 3. Adding a directory for proxied objects to be stored.
+# 4. Adding the entrypoint scripts found in the base NGINX OSS Docker image with a modified version that explicitly sets resolvers.
 
-ENTRYPOINT ["/docker-entrypoint.sh"]
+# Download your NGINX license certificate and key from the F5 customer portal (https://account.f5.com) and copy it to the build context
```
4141done, line 33: I wouldn't be sure what where I need to put my files based on this comment. I know we supply build context as the last argument to `docker build` so does that mean that if I supply `.` then this file expects my license files to be in the current dir?

Dockerfile.oss
```diff
@@ -1,63 +1,46 @@
...
-        pgp.mit.edu \
-    ; do \
-        echo "Fetching GPG key $NGINX_GPGKEY from $server"; \
-        gpg1 --keyserver "$server" --keyserver-options timeout=10 --recv-keys "$NGINX_GPGKEY" && found=yes && break; \
-    done; \
-    test -z "$found" && echo >&2 "error: failed to fetch GPG key $NGINX_GPGKEY" && exit 1; \
-    gpg1 --export "$NGINX_GPGKEY" > "$NGINX_GPGKEY_PATH" ; \
-    rm -rf "$GNUPGHOME"; \
-    apt-get remove --purge --auto-remove -y gnupg1 && rm -rf /var/lib/apt/lists/*; \
-    echo "deb [signed-by=/etc/apt/keyrings/nginx-archive-keyring.gpg] https://nginx.org/packages/mainline/debian/ $(echo $PKG_RELEASE | cut -f2 -d~) nginx" >> /etc/apt/sources.list.d/nginx.list; \
```
4141done, line 49: Any particular reason we're removing this? Did the official nginx image GPG keys get updated?

4141done (commented):
This all makes sense.  One question and one small suggestion.  I'm going to run some tests with this then approve.

Dockerfile.buildkit.plus
alessfg: You would pass the location via the `src` parameter in `--secret id=nginx-crt,src=plus/etc/ssl/nginx/nginx-repo.crt`

Dockerfile.oss
alessfg: They did! All three keys (including the new two ones) are now part of the official image and the two new keys do not expire (in theory).

4141done (approved):
Sorry for the delay!  Manually tested!  Ship it!

PR #349, opened 2024-12-11
feat: support PUT and DELETE

ATGardner:
### Proposed changes
copied changes from https://github.com/nginxinc/nginx-s3-gateway/pull/125/files (thanks for the good direction)

Describe the use case and detail of the change. If this PR addresses an issue on GitHub, make sure to include a link to that issue using one of the [supported keywords](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) in this PR description (not in the title of the PR).

### Checklist

- [x] I have signed the [F5 Contributor License Agreement (CLA)](https://github.com/f5/.github/blob/main/CLA/cla-markdown.md).
- [ ] If applicable, I have added tests that prove my fix is effective or that my feature works.
- [ ] If applicable, I have checked that any relevant tests pass after adding my changes.
- [ ] I have updated any relevant documentation (e.g. [`README.md`](/README.md)).

### DIFF
--- a/common/docker-entrypoint.sh
+++ b/common/docker-entrypoint.sh
@@ -57,11 +57,11 @@ export CORS_ENABLED="$(parseBoolean "${CORS_ENABLED}")"
 # is not normally used as part of the gateway. The following variable
 # defines the set of acceptable headers.
 if [ "${CORS_ENABLED}" == "1" ]; then
-  export LIMIT_METHODS_TO="GET HEAD OPTIONS"
-  export LIMIT_METHODS_TO_CSV="GET, HEAD, OPTIONS"
+  export LIMIT_METHODS_TO="GET HEAD OPTIONS PUT DELETE"
+  export LIMIT_METHODS_TO_CSV="GET, HEAD, OPTIONS, PUT, DELETE"
 else
-  export LIMIT_METHODS_TO="GET HEAD"
-  export LIMIT_METHODS_TO_CSV="GET, HEAD"
+  export LIMIT_METHODS_TO="GET HEAD PUT DELETE"
+  export LIMIT_METHODS_TO_CSV="GET, HEAD, PUT, DELETE"
 fi
 
 if [ -z "${CORS_ALLOWED_ORIGIN+x}" ]; then
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -366,13 +366,6 @@ function _s3DirQueryParams(uriPath, method) {
  * @param r {NginxHTTPRequest} HTTP request object
  */
 function redirectToS3(r) {
-    // This is a read-only S3 gateway, so we do not support any other methods
-    if (!(r.method === 'GET' || r.method === 'HEAD')) {
-        utils.debug_log(r, 'Invalid method requested: ' + r.method);
-        r.internalRedirect("@error405");
-        return;
-    }
-
     const uriPath = r.variables.uri_path;
     const isDirectoryListing = ALLOW_LISTING && _isDirectory(uriPath);
 


ATGardner:

dekobon:
Thank you for this, @ATGardner — for picking up a thread that had gone cold, for signing the CLA, and for being upfront about where the approach came from. I'm sorry it sat this long without a review. That's on us, not on you.

We've decided not to add write support to the gateway. The full reasoning is in #359, which I closed with a detailed explanation, so I won't repeat all of it here. The short version: the gateway holds one set of credentials and has no concept of client identity, so read-only is what keeps that arrangement coherent.

But I owe you more than "we decided no," because we did actually evaluate this, and **we found things that matter to you directly if you're running this patch on a fork.** That's the part I'd most like you to take away from this thread.

### The payload signing cliff

This is the one that would have hurt. SigV4 requires `x-amz-content-sha256` to hash the request body, and `awsHeaderPayloadHash` computes that from `$request_body`. nginx only populates that variable while the body fits in its memory buffers — past that it spools to a temp file and `$request_body` is **empty**. I measured it against the current image:

```
PUT    1024 bytes -> request_body_len=1024  sha256=2edc986847e209b4
PUT    8192 bytes -> request_body_len=8192  sha256=dd4e673052093276
PUT   16384 bytes -> request_body_len=0     sha256=e3b0c44298fc1c14   <- sha256("")
PUT   32768 bytes -> request_body_len=0     sha256=e3b0c44298fc1c14
PUT  524288 bytes -> request_body_len=0     sha256=e3b0c44298fc1c14
```

So uploads under ~8KB succeed, and uploads past the buffer sign the hash of an empty string while sending real bytes — S3 rejects those with `SignatureDoesNotMatch`. Small test files work; real files silently do not. That boundary is invisible from the diff, which is exactly why I don't consider this a mistake on your part.

There is a clean fix, for what it's worth: AWS supports the literal `UNSIGNED-PAYLOAD` as the `x-amz-content-sha256` value for single-chunk transfers, which skips body hashing entirely. Over TLS that's what most SDKs do for streaming uploads. If you are maintaining a fork, that change alone will get you past this.

### Three other things the surrounding config does

None of these are visible from a methods-allowlist diff either:

- **Failures arrive as 404.** `s3_location_common.conf` maps `403` to `404` via `error_page` — deliberate anti-reconnaissance for reads. It means a rejected upload surfaces as "404 Not Found" on a file you just wrote, with the real reason discarded.
- **1MB ceiling.** No `client_max_body_size` is set anywhere, so nginx's 1MB default applies. Anything larger gets a `413` before signing is even reached.
- **The cache doesn't learn about writes.** `proxy_cache_methods GET HEAD` means a `PUT` bypasses the cache without invalidating it, so a read right after your own write serves the stale object for up to `PROXY_CACHE_VALID_OK` (1h by default). Fixing that needs `proxy_cache_purge`, which is NGINX Plus only.

### On the patch itself

It also no longer applies — `common/docker-entrypoint.sh` was since split into `common/docker-entrypoint.d/`, so `LIMIT_METHODS_TO` now lives in `01-set-defaults.envsh`. Worth knowing if you rebase your fork.

### If you want to pursue this

I'm not closing the door on the idea, only on this diff. A version we could seriously consider would need: an opt-in env var defaulting to off, `UNSIGNED-PAYLOAD` signing with TLS enforced, a deliberate body-size limit, a separate error policy for write methods so failures are diagnosable, cache invalidation (or an explicit refusal to run writes with caching enabled), and integration tests covering that size boundary. #496 would need fixing first as well, since it's an open gap in the method restrictions that only becomes dangerous once writes are possible.

That's a design discussion rather than a patch, and if you'd like to have it, please do open an issue — you clearly have a real use case, and I'd rather hear it than have you carry a fork with the sharp edges above. Thanks again for the contribution and for your patience with our response time.

PR #447, opened 2025-09-03
feat: Move server_variables.conf to templates for read-only fs

vadim-tivikov:
### Proposed changes

This change enables the nginx-s3-gateway container to run with a read-only root filesystem (`readOnlyRootFilesystem: true`), which is a common security best practice in environments like Kubernetes.

### The Problem

When running a container with a read-only root filesystem, any directory that requires write access at runtime must be mounted as a writable volume (e.g., an `emptyDir`). The NGINX entrypoint script renders configuration templates from `/etc/nginx/templates/` into `/etc/nginx/conf.d/`. Therefore, `/etc/nginx/conf.d/` must be a writable volume.

However, the current image stores a critical configuration file, `gateway/server_variables.conf`, directly inside `/etc/nginx/conf.d/` in the image filesystem. When an `emptyDir` volume is mounted at `/etc/nginx/conf.d/`, it hides the original `server_variables.conf` file. This causes NGINX to fail on startup because the `include` directive for this file cannot find it.

### The Solution

This pull request resolves the issue by treating `server_variables.conf` as a template. It moves the file from:
`etc/nginx/conf.d/gateway/server_variables.conf`
to
`etc/nginx/templates/gateway/server_variables.conf.template`

By doing this, the standard container entrypoint script (`20-envsubst-on-templates.sh`) will now process this file and copy it into the writable `/etc/nginx/conf.d/gateway/` directory at container startup. This ensures the configuration is always present, even when `/etc/nginx/conf.d/` is a mounted volume.

This change is non-breaking for existing users and makes the S3 gateway compatible with modern, secure deployment patterns.

### Example Pod Manifest

Here is a simplified Pod manifest demonstrating how this change allows the container to run securely with a read-only root filesystem.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-s3-gateway-secure-pod
spec:
  automountServiceAccountToken: false
  securityContext:
    runAsUser: 101
    runAsGroup: 101
    fsGroup: 101
    runAsNonRoot: true
    seccompProfile:
      type: RuntimeDefault
  containers:
  - name: nginx
    image: nginxinc/nginx-s3-gateway:unprivileged-oss
    ports:
    - name: http
      containerPort: 8080
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      readOnlyRootFilesystem: true
    volumeMounts:
    - mountPath: /var/cache/nginx/
      name: nginx-cache
    - mountPath: /tmp/
      name: tmp
    - mountPath: /etc/nginx/conf.d/
      name: nginx-confd
  volumes:
  - name: nginx-cache
    emptyDir: {}
  - name: tmp
    emptyDir: {}
  - name: nginx-confd
    emptyDir: {}
```

### Checklist

- [x] I have signed the [F5 Contributor License Agreement (CLA)](https://github.com/f5/f5-cla/blob/main/docs/f5_cla.md).


vadim-tivikov:

dekobon (approved):
LGTM

PR #521, opened 2026-04-21
fix(conf): run auth_request for /index.html locations too

### LINKED ISSUE #518: `location ~ /index.html$` bypasses `auth_request` - TypeError with dynamic credentials
### Bug Overview

The `location ~ /index.html$` regex in `default.conf.template` does not have an `auth_request` directive. Nginx regex locations take priority over `location /`, so any request ending in `/index.html` skips credential fetching entirely. When using dynamic credentials (Pod Identity, IRSA, IMDS, or ECS task role), `readCredentials()` returns `undefined` and the request fails with a TypeError:

```
[error] js exception: TypeError: cannot get property "sessionToken" of undefined
    at signatureV4 (/etc/nginx/include/awssig4.js:51)
    at s3auth (/etc/nginx/include/s3gateway.js:209)
```

This affects every page served through `PROVIDE_INDEX_PAGE`, not just `/index.html` but also `/docs/index.html`, `/app/nested/index.html`, etc.

`auth_request /aws/credentials/retrieve;` only exists in `location /` (line 97 of `default.conf.template`). The `location ~ /index.html$` (line 249) takes priority for any path ending in `/index.html`. When the regex matches, no `auth_request` fires, so `fetchCredentials` never runs and `/tmp/credentials.json` is never written. Then `readCredentials()` returns `undefined`, and `signatureV4()` throws a TypeError accessing `credentials.sessionToken`.

The `@named` locations also lack `auth_request` but that's fine since they are only reachable via `internalRedirect` from `location /` where `auth_request` has already run. The regex location is the only directly-matchable location that is missing it.

When `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are set as environment variables, `readCredentials()` returns them from `process.env` directly and never touches the file cache or `auth_request` path. The bug only occurs with dynamic credential sources where `fetchCredentials` needs to run first.

This is a separate bug from #410 and #414, but tends to surface alongside them for Pod Identity users. Those issues deal with the correct credential source not being selected (missing `env` directives in `nginx.conf`). This issue is about `fetchCredentials` never being called at all for `index.html` paths, even after the `env` directives are correctly configured.

### Expected Behavior

Requests to paths ending in `/index.html` should fetch credentials before attempting to sign the S3 request, the same as any other path.

### Steps to Reproduce the Bug

1. Deploy the OSS gateway on EKS with Pod Identity (no static `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`).
2. Set `PROVIDE_INDEX_PAGE=true`.
3. Upload content to S3 at a path like `/docs/index.html`.
4. Request any `/path/index.html` directly on a fresh pod, before any other request has warmed the credential cache.

### Environment Details

- **Version of the S3 container used:** `ghcr.io/nginx/nginx-s3-gateway/nginx-oss-s3-gateway:unprivileged-oss-20260413`
- **Version of NGINX Open Source or NGINX Plus:** OSS
- **Target deployment platforms:** AWS EKS
- **S3 backend implementation:** AWS
- **Authentication method:** EKS Pod Identity

### Additional Context

**Suggested fix (preferred):** Move `auth_request /aws/credentials/retrieve;` to the `server` block level (via `server_variables.conf.template`), then add `auth_request off;` to locations that should not trigger it (`/health`, `/soap`, `/aws/credentials/retrieve`, and the `@named` locations). This way any new directly-matchable locations added in the future will automatically get credential fetching.

**Suggested fix (minimal):** Add `auth_request /aws/credentials/retrieve;` to `location ~ /index.html$`.

**Secondary issue:** `awscredentials.js` line 39 defines `const NOW = new Date()` at module scope. This value is used for both S3 signature timestamps and the credential expiration check (line 273):

```js
if (NOW.getTime() < exp) {
    r.return(200);  // skip refresh
    return;
}
```

This works today because njs re-evaluates module-scope code per request context, so `NOW` gets a fresh value each time. But the correctness depends on this njs implementation detail. If context reuse behavior changes (e.g. `js_context_reuse`), credentials would silently stop refreshing. The expiration check should use `new Date()` explicitly:

```js
if (new Date().getTime() < exp) {
```

SAY-5:
## Summary

Fixes #518.

nginx regex locations beat the `location /` prefix match, so direct requests to any `*/index.html` path skipped the `auth_request /aws/credentials/retrieve` sub-request that only fires inside `location /`. When `PROVIDE_INDEX_PAGE=true` is combined with dynamic credential sources (Pod Identity / IRSA / IMDS / ECS task role), `fetchCredentials` never runs, `/tmp/credentials.json` stays empty, and `signatureV4` crashes:

```
[error] js exception: TypeError: cannot get property "sessionToken" of undefined
    at signatureV4 (/etc/nginx/include/awssig4.js:51)
    at s3auth (/etc/nginx/include/s3gateway.js:209)
```

Affects every page served through `PROVIDE_INDEX_PAGE`, not just the root `/index.html`.

## Fix

Add `auth_request /aws/credentials/retrieve;` to the regex location so credentials are refreshed for every user-reachable path. Named locations (`@s3`, `@s3PreListing`, `@error404`) still inherit through `internalRedirect` and don't need the directive. Static-key deployments are untouched because `readCredentials` short-circuits on `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` env vars before the auth_request matters.

## Test plan

- [x] Reviewed the auth_request flow to confirm `@named` locations stay reachable through `internalRedirect` and don't need a direct auth_request
- [ ] CI integration tests in `test/integration/` cover static-key and dynamic-credential setups; would value a review of whether a specific PROVIDE_INDEX_PAGE + Pod-Identity integration test is worth adding as a follow-up


dekobon (approved):
LGTM

PR #523, opened 2026-05-03
fix: prevent hasExtension matching non terminal sub-paths

### LINKED ISSUE #522: Append slash configuration fails when non terminal path segment includes '.'
### Bug Overview

The regex for `trailsslashControl` incorrectly classifies the directory `/foo.bar/bar/baz` as a file with extension, causing the `APPEND_SLASH_FOR_POSSIBLE_DIRECTORY` configuration to be ignored.

### Expected Behavior

When starting the gateway with `APPEND_SLASH_FOR_POSSIBLE_DIRECTORY` active, only the final segment of the URL path should be considered to identify if path refers to a directory but due to an improper regex used in `s3gateway.js` this is not true. 

Here's a breakdown of use-cases and expected behaviours:

| Path | Current triggers append slash | Expected triggers append slash |
|--------|--------|--------|
| /foo/bar/baz | Yes | Yes |
| /foo/bar.bar/baz.baz | No | No |
| /foo/bar/baz.baz | Yes | Yes |
| /foo.foo/bar/baz | No | Yes |
| /foo/bar.bar/baz | No | Yes |
| /foo/bar/.baz | Yes | No |
| /foo/bar/baz. | Yes | No |

### Steps to Reproduce the Bug

Start the gateway and create in the bucket a directory structure with path `/foo.foo/bar` and place `index.html` in the `bar` sub directory, then attempt to resolve `GET /foo.foo/bar`. The gateway should resolve the `index.html`, but it will return a 404 instead.

### Environment Details

- Version of the S3 container used: `pgsty/minio:RELEASE.2026-03-25T00-00-00Z`
- Commit/Tag (if building the NGINX S3 gateway from source): `98359a187ee212d6c78e6eb27a772b8c6c3cbe31`
- Version of NGINX Open Source or NGINX Plus: OSS
- Version of NGINX JavaScript: from  `98359a187ee212d6c78e6eb27a772b8c6c3cbe31`
- Target deployment platforms [e.g. AWS/GCP/local cluster/etc...]: all
- S3 backend implementation: [e.g. AWS, Ceph, NetApp StorageGrid, etc...]: all
- Authentication method: [e.g. IAM, IAM with Fargate, IAM with K8S, AWS Credentials, etc...]: all

### Additional Context

The regex can be improved to cover for the misbehaviour as follows:

`/(\/[^\/]*)*(\/[^\/]*\.[^.\/]*)$/`

This regex works as intendend because:

- The group `(\/[^\/]*)*` matches all sub-segments of the path that are not terminal
- The group `(\/[^\/]*\.[^.\/]*)` matches the last segment (`\/`) which includes a `.` character that is either followed or not by any other character.

fburato:
### Proposed changes

Update the `s3gateway.js` regular expression for the `hasExtension` check to prevent it from detecting non terminal sub-path as potential part of an extension. This translates in the `APPEND_SLASH_FOR_POSSIBLE_DIRECTORY` config enabled behaviour to trigger for paths like `/foo.foo/bar`.

Fixes #522.

### Checklist

- [x] I have signed the [F5 Contributor License Agreement (CLA)](https://github.com/f5/f5-cla/blob/main/docs/f5_cla.md).

### DIFF
--- a/common/etc/nginx/include/s3gateway.js
+++ b/common/etc/nginx/include/s3gateway.js
@@ -398,7 +398,7 @@ function trailslashControl(r) {
         // consider the uri without query params or anchors
         const path = r.variables.uri_path.split(/[?#]/)[0];
 
-        const hasExtension = /\/[^.\/]+\.[^.]+$/;
+        const hasExtension = /(\/[^\/]*)*(\/[^\/]*\.[^.\/]*)$/;
         if (!hasExtension.test(path)  && !_isDirectory(path)){
             return r.internalRedirect("@trailslash");
         }


fburato:

dekobon:
Thank you @fburato for the fix and for the excellent analysis in #522! This change has been merged to `main` via #546, which carried your commit with authorship preserved (co-author credit on the squashed commit d6c0fed). That PR also added unit tests covering the new extension-detection behavior, plus two follow-ups found while reviewing: the redundant leading group in the regex was removed (it caused quadratic backtracking on long request URIs), and `trailslashControl` now decodes the path before classification so percent-encoded dots (e.g. `%2E`) are handled consistently. Closing this PR as superseded — thanks again for the contribution!

PR #532, opened 2026-06-08
ci: fetch secrets from Azure

thresheek:
### Proposed changes

Instead of storing secrets on GH, let's move them to a more manageable service.


thresheek:
Fixed commit message to make Lint/Validate job less unhappy

oxpa (approved):

PR #538, opened 2026-08-17
fix: do not cache a non-200 EKS Pod Identity agent response as credentials

### LINKED ISSUE #537: EKS Pod Identity: a non-200 agent response is cached as credentials, gateway then serves 404 for every object until the next successful refresh
## Describe the bug

`_fetchEKSPodIdentityCredentials` in `common/etc/nginx/include/awscredentials.js` does not check `resp.ok` before parsing the response:

https://github.com/nginx/nginx-s3-gateway/blob/1e0e5e7a7d35ef7945e7403866b7685e46a1d6b6/common/etc/nginx/include/awscredentials.js#L403-L417

When the EKS Pod Identity agent answers non-200 (token rotation blip, agent restart or throttling), the error body is parsed as if it were credentials, so the gateway caches `{accessKeyId: undefined, secretAccessKey: undefined, ...}` via `writeCredentials`. Every S3 request is then signed with those undefined values, S3 rejects them with 400, and because the gateway intentionally sanitizes upstream errors to 404 (`proxy_intercept_errors`), the whole site serves 404 for objects that exist until a later refresh succeeds.

The sibling function `_fetchEcsRoleCredentials` already guards exactly this case:

https://github.com/nginx/nginx-s3-gateway/blob/1e0e5e7a7d35ef7945e7403866b7685e46a1d6b6/common/etc/nginx/include/awscredentials.js#L338-L343

so this looks like an oversight in #252 rather than a design choice. (`_fetchEC2RoleCredentials` has the same unguarded pattern on its three fetches, for what it's worth.)

## To reproduce

1. Run the gateway on EKS with Pod Identity (`AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` set), serving any bucket.
2. Make the agent return an error for a refresh window (e.g. restart the `eks-pod-identity-agent` pod, or point the token file at a rotated/invalid token).
3. Once the cached credentials pass their refresh window, the next `fetchCredentials` stores the error response as credentials.
4. Every request now returns 404; S3 access logs / `upstream_status` show 400 (`InvalidAccessKeyId` family) behind them.
5. The gateway recovers on its own once the agent answers 200 again.

We hit this in production behind report-hosting sites: intermittent windows where existing objects 404, self-healing minutes later, with `upstream_status: 400` logged for each. Confirmed on `unprivileged-oss-20260810`.

## Expected behavior

A failed agent response should make the refresh fail (as the ECS path does), keeping the previously cached credentials, which are normally still valid through an agent blip, instead of poisoning the cache with undefined values.

## Your environment

- Version: `ghcr.io/nginx/nginx-s3-gateway/nginx-oss-s3-gateway:unprivileged-oss-20260810`
- EKS, credentials via `aws_eks_pod_identity_association`
- S3 (real, eu-west-1), virtual-host style

Happy to send the one-line fix (mirror the ECS `resp.ok` guard) plus an error-path unit test as a PR.

tomknee-accelins:
### Proposed changes

Fixes #537.

`_fetchEKSPodIdentityCredentials` calls `resp.json()` without checking `resp.ok`. When the EKS Pod Identity agent answers non-200 (token rotation blip, agent restart, throttling), the error body is parsed as if it were credentials, `writeCredentials` caches `{accessKeyId: undefined, ...}`, and every S3 request is then signed with those values. S3 rejects them with 400, which the gateway sanitizes to 404, so the whole site serves 404 for existing objects until a later refresh succeeds. Details and production evidence in the linked issue.

This adds the same `resp.ok` guard `_fetchEcsRoleCredentials` already has, so a failed refresh throws (surfacing as the existing 500 path in `fetchCredentials`) and the previously cached credentials, normally still valid through an agent blip, are kept instead of being overwritten.

Also adds an error-path unit test: a mocked non-200 agent response must produce a 500 from `fetchCredentials` and must never have its body parsed as credentials. Verified with the repo's unit harness against the published image (`njs -m -p /etc/nginx /var/tmp/awscredentials_test.js` in `unprivileged-oss-20260810`): all existing tests still pass and the new test passes with the guard, failing without it.

### Checklist

- [ ] I have signed the [F5 Contributor License Agreement (CLA)](https://github.com/f5/f5-cla/blob/main/docs/f5_cla.md).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/unit/awscredentials_test.js
+++ b/test/unit/awscredentials_test.js
@@ -372,10 +372,63 @@ async function testEKSPodIdentityCredentialRetrieval() {
     }
 }
 
+async function testEKSPodIdentityCredentialRetrievalNon200Response() {
+    printHeader('testEKSPodIdentityCredentialRetrievalNon200Response');
+    if ('AWS_ACCESS_KEY_ID' in process.env) {
+        delete process.env['AWS_ACCESS_KEY_ID'];
+    }
+    if ('AWS_CONTAINER_CREDENTIALS_RELATIVE_URI' in process.env) {
+        delete process.env['AWS_CONTAINER_CREDENTIALS_RELATIVE_URI'];
+    }
+    if ('AWS_WEB_IDENTITY_TOKEN_FILE' in process.env) {
+        delete process.env['AWS_WEB_IDENTITY_TOKEN_FILE'];
+    }
+    var tempDir = (process.env['TMPDIR'] ? process.env['TMPDIR'] : '/tmp');
+    var uniqId = `${new Date().getTime()}-${Math.floor(Math.random()*101)}`;
+    var tempFile = `${tempDir}/credentials-unit-test-${uniqId}.json`;
+    var testToken = 'A_TOKEN';
+    fs.writeFileSync(tempFile, testToken);
+    process.env['AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE'] = tempFile;
+    globalThis.errorBodyParsedAsCredentials = false;
+    globalThis.ngx.fetch = function(url, options) {
+        console.log(' fetching eks pod identity mock error response');
+        return Promise.resolve({
+            ok: false,
+            status: 503,
+            json: function() {
+                // A non-200 body must never be parsed as credentials.
+                globalThis.errorBodyParsedAsCredentials = true;
+                return Promise.resolve({
+                    message: 'Service Unavailable',
+                });
+            },
+        });
+    };
+    var returnedCode = null;
+    var r = {
+        log: function(msg) {
+            console.log(msg);
+        },
+        return: function(code) {
+            returnedCode = code;
+        },
+    };
+
+    await awscred.fetchCredentials(r);
+
+    if (globalThis.errorBodyParsedAsCredentials) {
+        throw 'A non-200 EKS Pod Identity agent response was parsed as credentials.';
+    }
+    if (returnedCode !== 500) {
+        throw 'Expected the credentials fetch to fail with 500, got: ' + returnedCode;
+    }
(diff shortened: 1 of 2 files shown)


test/unit/awscredentials_test.js
```diff
@@ -372,10 +372,63 @@ async function testEKSPodIdentityCredentialRetrieval() {
...
+    }
+    var tempDir = (process.env['TMPDIR'] ? process.env['TMPDIR'] : '/tmp');
+    var uniqId = `${new Date().getTime()}-${Math.floor(Math.random()*101)}`;
+    var tempFile = `${tempDir}/credentials-unit-test-${uniqId}.json`;
+    var testToken = 'A_TOKEN';
+    fs.writeFileSync(tempFile, testToken);
+    process.env['AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE'] = tempFile;
+    globalThis.errorBodyParsedAsCredentials = false;
+    globalThis.ngx.fetch = function(url, options) {
+        console.log(' fetching eks pod identity mock error response');
```
dekobon: Use `utils.debug_log` instead of `console.log.

test/unit/awscredentials_test.js
```diff
@@ -372,10 +372,63 @@ async function testEKSPodIdentityCredentialRetrieval() {
...
+                return Promise.resolve({
+                    message: 'Service Unavailable',
+                });
+            },
+        });
+    };
+    var returnedCode = null;
+    var r = {
+        log: function(msg) {
+            console.log(msg);
```
dekobon: Use `utils.debug_log` instead of `console.log.

